PR #16, opened 2021-03-19
ENH: Nipype-less implementation (refactor)

oesteban:
Setting some foundations for the code:

- The DWI object to encapsulate inputs (please let me know if it encloses all we will possibly need).
- Stubb to start building the core of the implementation, defining some sort of API.

UPDATE:

This is how it looks right now:
```Python
>>> from emc import dmri
>>> from emc.estimator import EddyMotionEstimator
>>> data = dmri.load(
...     "ds001771/sub-36/dwi/sub-36_acq-AP_dwi.nii.gz",
...     gradients_file="rasb.tsv",
...     b0_file="sub-36_acq-PA_dwi_b0_mcf_ref.nii.gz",
...     brainmask_file="mask.nii.gz"
...     )
>>> EddyMotionEstimator.fit(data, model="b0")
Processing b-index <19> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpedeh8iiv>
Processing b-index <35> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmppb9tb2cz>
Processing b-index <54> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpe9yk0dez>
Processing b-index <98> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp0scj6vxi>
Processing b-index <65> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpgeah9t8a>
Processing b-index <0> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpmjqioqk4>
Processing b-index <47> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp3vqbgpvm>
Processing b-index <60> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp2exvtr9t>
Processing b-index <45> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmphljr6ty6>
Processing b-index <27> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp64ygidbx>
Processing b-index <39> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpqf8xfcf6>
Processing b-index <97> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpw6f574ft>
Processing b-index <24> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmphrds6nq9>
Processing b-index <87> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpjmja70vg>
Processing b-index <76> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmplflxbnov>
Processing b-index <16> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpezw74z4b>
Processing b-index <25> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpfx5a80rl>
Processing b-index <71> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp0mp13_we>
Processing b-index <22> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp6eqn7q0o>
Processing b-index <10> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp44f58w84>
Processing b-index <15> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpixktyl6w>
Processing b-index <53> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp2u3ayj_9>
Processing b-index <48> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpzbsozmz7>
Processing b-index <61> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp7034ovrf>
Processing b-index <20> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpmvscnyvg>
Processing b-index <46> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpczdgmj4z>
Processing b-index <2> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpvrmfs29_>
Processing b-index <90> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpi28jxzkz>
Processing b-index <18> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp_j8qvw2h>
Processing b-index <69> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmp_ufs8o7v>
Processing b-index <40> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpf2oom5r2>
Processing b-index <6> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmpgjescu1p>
Processing b-index <28> in </var/folders/l9/0lkn3g4s27bgkk75n6jj778r0000gp/T/tmppk70ev6x>
31%|█████████████████                              | 32/102 [11:24<32:56, 28.24s/it]
```

### DIFF
--- a/emc/dmri.py
+++ b/emc/dmri.py
@@ -35,7 +35,7 @@ class DWI:
     """List of linear matrices that bring DWIs (i.e., no b=0) into alignment."""
     fieldmap = attr.ib(default=None, repr=_data_repr)
     """A 3D displacements field to unwarp susceptibility distortions."""
-    _filepath = attr.ib(default=None)
+    _filepath = attr.ib(default=None, repr=False)
     """A path to an HDF5 file to store the whole dataset."""
 
     def __len__(self):
@@ -80,7 +80,7 @@ class DWI:
             xform = Affine(matrix=affine, reference=reference)
 
         # read original DWI data & b-vector
-        with h5py.File(self.filepath, "r") as in_file:
+        with h5py.File(self._filepath, "r") as in_file:
             root = in_file["/0"]
             dwframe = root["dataobj"][..., index]
             bvec = root["gradients"][:3, index]
@@ -133,7 +133,7 @@ class DWI:
         return retval
 
 
-def load(filename, gradients_file, b0_file=None, fmap_file=None):
+def load(filename, gradients_file=None, b0_file=None, fmap_file=None):
     """Load DWI data."""
     import nibabel as nb
 
@@ -141,11 +141,14 @@ def load(filename, gradients_file, b0_file=None, fmap_file=None):
     if filename.name.endswith(".h5"):
         return DWI.from_filename(filename)
 
+    if not gradients_file:
+        raise RuntimeError("A gradients file is necessary")
+
     img = nb.as_closest_canonical(nb.load(filename))
     retval = DWI(
         affine=img.affine,
     )
-    grad = np.loadtxt(gradients_file).T
+    grad = np.loadtxt(gradients_file, dtype="float32").T
     gradmsk = grad[-1] > 50
     retval.gradients = grad[..., gradmsk]
     retval.dataobj = img.get_fdata(dtype="float32")[..., gradmsk]
--- a/emc/dmri.py
+++ b/emc/dmri.py
@@ -1,9 +1,11 @@
 """Representing data in hard-disk and memory."""
+from pathlib import Path
 from collections import namedtuple
+from tempfile import mkstemp
 import attr
 import numpy as np
 import h5py
-from pathlib import Path
+import nibabel as nb
 from nitransforms.linear import Affine
 
 
@@ -72,6 +74,9 @@ class DWI:
             shape=self.dataobj.shape[:3], affine=self.affine
         )
 
+        if not self._filepath:
+            self.to_filename(mkstemp(suffix=".h5")[1])
+
         # create a nitransforms object
         if self.fieldmap:
             # compose fieldmap into transform
@@ -85,15 +90,19 @@ class DWI:
             dwframe = root["dataobj"][..., index]
             bvec = root["gradients"][:3, index]
 
+        dwmoving = nb.Nifti1Image(dwframe, self.affine, None)
+
         # resample and update orientation at index
-        self.dataobj[..., index] = xform.apply(dwframe)
+        self.dataobj[..., index] = np.asanyarray(
+            xform.apply(dwmoving).dataobj, dtype=self.dataobj.dtype
+        )
 
         # invert transform transform b-vector and origin
-        r_bvec = ~xform.apply([bvec, (0.0, 0.0, 0.0)])
+        r_bvec = (~xform).map([bvec, (0.0, 0.0, 0.0)])
         # Reset b-vector's origin
         new_bvec = r_bvec[1] - r_bvec[0]
         # Normalize and update
-        self.gradients[:3, index] = new_bvec / np.norm(new_bvec)
+        self.gradients[:3, index] = new_bvec / np.linalg.norm(new_bvec)
 
         # update transform
         if self.em_affines is None:
@@ -133,10 +142,10 @@ class DWI:
         return retval
 
 
-def load(filename, gradients_file=None, b0_file=None, fmap_file=None):
+def load(
+    filename, gradients_file=None, b0_file=None, brainmask_file=None, fmap_file=None
+):
     """Load DWI data."""
-    import nibabel as nb
-
     filename = Path(filename)
     if filename.name.endswith(".h5"):
         return DWI.from_filename(filename)
@@ -156,6 +165,10 @@ def load(filename, gradients_file=None, b0_file=None, fmap_file=None):
     if b0_file:
         b0img = nb.as_closest_canonical(nb.load(b0_file))
         retval.bzero = np.asanyarray(b0img.dataobj)
+
+    if brainmask_file:
+        mask = nb.as_closest_canonical(nb.load(brainmask_file))
+        retval.brainmask = np.asanyarray(mask.dataobj)
 
     if fmap_file:
         fmapimg = nb.as_closest_canonical(nb.load(fmap_file))
--- a/emc/dmri.py
+++ b/emc/dmri.py
@@ -37,7 +37,7 @@ class DWI:
     """List of linear matrices that bring DWIs (i.e., no b=0) into alignment."""
     fieldmap = attr.ib(default=None, repr=_data_repr)
     """A 3D displacements field to unwarp susceptibility distortions."""
-    _filepath = attr.ib(default=None, repr=False)
+    _filepath = attr.ib(default=mkstemp(suffix=".h5")[1], repr=False)
     """A path to an HDF5 file to store the whole dataset."""
 
     def __len__(self):
@@ -74,15 +74,15 @@ class DWI:
             shape=self.dataobj.shape[:3], affine=self.affine
         )
 
-        if not self._filepath:
-            self.to_filename(mkstemp(suffix=".h5")[1])
-
         # create a nitransforms object
         if self.fieldmap:
             # compose fieldmap into transform
             raise NotImplementedError
         else:
             xform = Affine(matrix=affine, reference=reference)
+
+        if not Path(self._filepath).exists():
+            self.to_filename(self._filepath)
 
         # read original DWI data & b-vector
         with h5py.File(self._filepath, "r") as in_file:
@@ -131,7 +131,16 @@ class DWI:
                         f.name,
                         data=value,
                     )
-        self._filepath = filename
+
+    def to_nifti(self, filename):
+        """Write a NIfTI 1.0 file to disk."""
+        nii = nb.Nifti1Image(
+            self.dataobj,
+            self.affine,
+            None,
+        )
+        nii.header.set_xyzt("mm")
+        nii.to_filename(filename)
 
     @classmethod
     def from_filename(cls, filename):
--- a/emc/dmri.py
+++ b/emc/dmri.py
@@ -1,0 +1,96 @@
+"""Representing data in hard-disk and memory."""
+import attr
+import numpy as np
+import h5py
+from pathlib import Path
+
+
+@attr.s(slots=True)
+class DWI:
+    """Data representation structure for dMRI data."""
+
+    dataobj = attr.ib(default=None)
+    """A numpy ndarray object for the data array, without *b=0* volumes."""
+    affine = attr.ib(default=None)
+    """Best affine for RAS-to-voxel conversion of coordinates (NIfTI header)."""
+    brainmask = attr.ib(default=None)
+    """A boolean ndarray object containing a corresponding brainmask."""
+    bzero = attr.ib(default=None)
+    """
+    A *b=0* reference map, preferably obtained by some smart averaging.
+    If the :math:`B_0` fieldmap is set, this *b=0* reference map should also
+    be unwarped.
+    """
+    gradients = attr.ib(default=None)
+    """A 2D numpy array of the gradient table in RAS+B format."""
+    sampling = attr.ib(default=None)
+    """Sampling of q-space: single-, multi-shell or cartesian."""
+    em_affines = attr.ib(default=None)
+    """List of linear matrices that bring DWIs (i.e., no b=0) into alignment."""
+    fieldmap = attr.ib(default=None)
+    """A 3D displacements field to unwarp susceptibility distortions."""
+
+    def logo_split(self, index):
+        """
+        Produce one fold of LOGO (leave-one-gradient-out).
+
+        Parameters
+        ----------
+        index : :obj:`int`
+            Index of the DWI orientation to be left out in this fold.
+
+        Return
+        ------
+        (train_data, train_gradients) : :obj:`tuple`
+            Training DWI and corresponding gradients
+        (test_data, test_gradients) :obj:`tuple`
+            Test 3D map (one DWI orientation) and corresponding b-vector/value.
+
+        """
+        mask = np.zeros(self.gradients.shape[-1], dtype=bool)
+        mask[index] = True
+        mask = mask[np.newaxis, np.newaxis, np.newaxis, :]
+        return (
+            (self.dataobj[~mask], self.gradients[~mask]),
+            (self.dataobj[mask], self.gradients[mask]),
+        )
+
+    def to_filename(self, filename):
+        """Write an HDF5 file to disk."""
+        filename = Path(filename)
+        if not filename.endswith(".h5"):
+            filename = filename.parent / f"{filename.name}.h5"
+
+        with h5py.File(filename, "w") as out_file:
+            out_file.attrs["Format"] = "EMC/DWI"
+            out_file.attrs["Version"] = np.uint16(1)
+            root = out_file.create_group("/0")
+            root.attrs["Type"] = "dwi"
+            for f in attr.fields(self.__class__):
+                root.create_dataset(
+                    f.name,
+                    data=getattr(self, f.name),
+                )
+
+    @classmethod
+    def from_filename(cls, filename):
+        """Read an HDF5 file from disk."""
+        with h5py.File(filename, "w") as in_file:
+            root = in_file["/0"]
+            retval = cls(**{k: v for k, v in root.items()})
+        return retval
+
+
+def load(filename):
+    """Load DWI data."""
+    import nibabel as nb
+
+    filename = Path(filename)
+    if filename.name.endswith(".h5"):
+        return DWI.from_filename(filename)
+
+    img = nb.as_closest_canonical(nb.load(filename))
+    return DWI(
+        dataobj=img.get_fdata(dtype="float32"),
+        affine=img.affine,
+    )
--- a/emc/dmri.py
+++ b/emc/dmri.py
@@ -56,19 +56,31 @@ class DWI:
         Return
         ------
         (train_data, train_gradients) : :obj:`tuple`
-            Training DWI and corresponding gradients
+            Training DWI and corresponding gradients.
+            Training data/gradients come **from the updated dataset**.
         (test_data, test_gradients) :obj:`tuple`
             Test 3D map (one DWI orientation) and corresponding b-vector/value.
+            The test data/gradient come **from the original dataset**.
 
         """
+        if not Path(self._filepath).exists():
+            self.to_filename(self._filepath)
+
+        # read original DWI data & b-vector
+        with h5py.File(self._filepath, "r") as in_file:
+            root = in_file["/0"]
+            dwframe = np.asanyarray(root["dataobj"][..., index])
+            bframe = np.asanyarray(root["gradients"][..., index])
+
+        # if the size of the mask does not match data, cache is stale
         mask = np.zeros(len(self), dtype=bool)
         mask[index] = True
         return (
             (self.dataobj[..., ~mask], self.gradients[..., ~mask]),
-            (self.dataobj[..., mask], self.gradients[..., mask]),
+            (dwframe, bframe),
         )
 
-    def set_transform(self, index, affine, order=1):
+    def set_transform(self, index, affine, order=3):
         """Set an affine, and update data object and gradients."""
         reference = namedtuple("ImageGrid", ("shape", "affine"))(
             shape=self.dataobj.shape[:3], affine=self.affine
@@ -87,14 +99,15 @@ class DWI:
         # read original DWI data & b-vector
         with h5py.File(self._filepath, "r") as in_file:
             root = in_file["/0"]
-            dwframe = root["dataobj"][..., index]
-            bvec = root["gradients"][:3, index]
+            dwframe = np.asanyarray(root["dataobj"][..., index])
+            bvec = np.asanyarray(root["gradients"][:3, index])
 
         dwmoving = nb.Nifti1Image(dwframe, self.affine, None)
 
         # resample and update orientation at index
         self.dataobj[..., index] = np.asanyarray(
-            xform.apply(dwmoving).dataobj, dtype=self.dataobj.dtype
+            xform.apply(dwmoving, order=order).dataobj,
+            dtype=self.dataobj.dtype,
         )
 
         # invert transform transform b-vector and origin
@@ -106,7 +119,7 @@ class DWI:
 
         # update transform
         if self.em_affines is None:
-            self.em_affines = [np.eye(4)] * len(self)
+            self.em_affines = [None] * len(self)
 
         self.em_affines[index] = xform
 
@@ -139,7 +152,7 @@ class DWI:
             self.affine,
             None,
         )
-        nii.header.set_xyzt("mm")
+        nii.header.set_xyzt_units("mm")
         nii.to_filename(filename)
 
     @classmethod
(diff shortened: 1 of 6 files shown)


emc/dmri.py
```diff
@@ -0,0 +1,96 @@
...
+import numpy as np
+import h5py
+from pathlib import Path
+
+
+@attr.s(slots=True)
+class DWI:
+    """Data representation structure for dMRI data."""
+
+    dataobj = attr.ib(default=None)
```
arokem, line 12: Does this name follow a specific convention? Maybe better to call it `dwdata` to be more explicit that these are the diffusion weighted images?

emc/dmri.py
```diff
@@ -0,0 +1,96 @@
...
+    """A numpy ndarray object for the data array, without *b=0* volumes."""
```
arokem, line 13: Is it just me, or is it weird that these strings appear _after_ the attr they refer to?

emc/dmri.py
```diff
@@ -0,0 +1,96 @@
...
+    sampling = attr.ib(default=None)
+    """Sampling of q-space: single-, multi-shell or cartesian."""
+    em_affines = attr.ib(default=None)
+    """List of linear matrices that bring DWIs (i.e., no b=0) into alignment."""
+    fieldmap = attr.ib(default=None)
+    """A 3D displacements field to unwarp susceptibility distortions."""
+
+    def logo_split(self, index):
+        """
+        Produce one fold of LOGO (leave-one-gradient-out).
```
arokem, line 35: "LOGO". Love it!

emc/estimator.py
```diff
@@ -0,0 +1,52 @@
+"""A SHORELine-like algorithm for the realignment of dMRI data."""
+import attr
+import numpy as np
+from pathlib import Path
+
+
+@attr.s(slots=True)
+class EddyMotionEstimation:
```
arokem, line 8:
```suggestion
class EddyMotionEstimator:
```

emc/estimator.py
```diff
@@ -0,0 +1,52 @@
...
+    """Estimate rigid-body head-motion and distortions derived from eddy-currents."""
```
arokem, line 9:
```suggestion
    """Estimates rigid-body head-motion and distortions derived from eddy-currents."""
```

emc/estimator.py
```diff
@@ -0,0 +1,52 @@
...
+
+    datapath = attr.ib(default=None)
+    """Path to the input DWI dataset."""
+    gradients = attr.ib(default=None)
+    """A 2D numpy array of the gradient table in RAS+B format."""
```
arokem, line 14: Why is the data a path and the gradients an array? Maybe we could allow either of these as input for both of these?

emc/estimator.py
```diff
@@ -0,0 +1,52 @@
...
+    bzero = attr.ib(default=None)
+    """A *b=0* reference map, preferably obtained by some smart averaging."""
```
arokem, line 16: Is this a path or an array? I would advocate for either.

emc/estimator.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        loo_index = np.random.shuffle(range(len(X)))
+
+        self.transforms = [
+            _emc(X, index=i)
+            for i in loo_index
+        ]
+        raise NotImplementedError
+
+    def predict():
```
arokem, line 47: I think this might be more akin to the sklearn `transform` method of unsupervised estimators.

And then we can also add a `fit_transform` method that does both.

arokem (commented):
Looks good! One general point is that I think that this probably ends up looking more like sklearn unsupervised estimators. For example, note that the `predict` method takes no input. This means that it is probably more akin to the sklearn `transform` methods in unsupervised estimators. And we can add a `fit_transform` that does both. Other small comments spread throughout.

emc/dmri.py
dPys: Agreed. Best acronym I've heard so far this year.

emc/dmri.py
oesteban: nibabel gives access to the numpy array through `dataobj`, but it is not required, honestly.

emc/dmri.py
oesteban: It's a rejected PEP which sphinx honors (so these attributes will indeed appear documented)

emc/estimator.py
oesteban: I'm still thinking about this one, and go either one way or another (but not the way it is now, like you said).

If we want to enjoy the benefits of HDF5, just a path would be the best option. Otherwise, yes, we want to have all-arrays probably.

emc/estimator.py
oesteban: sounds good!

oesteban:
I hadn't assigned review because it felt premature - but this feedback is super-welcome. Thanks for your time, I will move to the better API tomorrow.

emc/estimator.py
oesteban: actually, with the new dwi object, the idea was to remove all of these.

oesteban:
@arokem I'm getting this:

```Python
>>> EddyMotionEstimator.fit(data)
Optimizing level 0 [max iter: 100]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-35-e34da2fc4915> in <module>
----> 1 EddyMotionEstimator.fit(data)

~/workspace/EddyMotionCorrection/emc/estimator.py in fit(dwdata, n_iter, align_kwargs, model, seed, **kwargs)
     99
    100                 init_affine = dwdata.em_affines[i] if dwdata.em_affines else np.eye(4)
--> 101                 xform = registration.optimize(
    102                     predicted,  # fixed
    103                     data_test[0],  # moving

~/.anaconda/lib/python3.8/site-packages/dipy/align/imaffine.py in optimize(***failed resolving arguments***)
   1099                 self.options['maxiter'] = max_iter
   1100
-> 1101             opt = Optimizer(self.metric.distance_and_gradient,
   1102                             self.params0,
   1103                             method=self.method, jac=True,

~/.anaconda/lib/python3.8/site-packages/dipy/core/optimize.py in __init__(self, fun, x0, args, method, jac, hess, hessp, bounds, constraints, tol, callback, options, evolution)
    127         else:
    128
--> 129             res = minimize(fun, x0, args, method, jac, hess, hessp, bounds,
    130                            constraints, tol, callback, options)
    131

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/_minimize.py in minimize(fun, x0, args, method, jac, hess, hessp, bounds, constraints, tol, callback, options)
    615                                   **options)
    616     elif meth == 'l-bfgs-b':
--> 617         return _minimize_lbfgsb(fun, x0, args, jac, bounds,
    618                                 callback=callback, **options)
    619     elif meth == 'tnc':

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/lbfgsb.py in _minimize_lbfgsb(fun, x0, args, jac, bounds, disp, maxcor, ftol, gtol, eps, maxfun, maxiter, iprint, callback, maxls, finite_diff_rel_step, **unknown_options)
    304             iprint = disp
    305
--> 306     sf = _prepare_scalar_function(fun, x0, jac=jac, args=args, epsilon=eps,
    307                                   bounds=new_bounds,
    308                                   finite_diff_rel_step=finite_diff_rel_step)

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/optimize.py in _prepare_scalar_function(fun, x0, jac, args, bounds, epsilon, finite_diff_rel_step, hess)
    259     # ScalarFunction caches. Reuse of fun(x) during grad
    260     # calculation reduces overall function evaluations.
--> 261     sf = ScalarFunction(fun, x0, args, grad, hess,
    262                         finite_diff_rel_step, bounds, epsilon=epsilon)
    263

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/_differentiable_functions.py in __init__(self, fun, x0, args, grad, hess, finite_diff_rel_step, finite_diff_bounds, epsilon)
     74
     75         self._update_fun_impl = update_fun
---> 76         self._update_fun()
     77
     78         # Gradient evaluation

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/_differentiable_functions.py in _update_fun(self)
    164     def _update_fun(self):
    165         if not self.f_updated:
--> 166             self._update_fun_impl()
    167             self.f_updated = True
    168

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/_differentiable_functions.py in update_fun()
     71
     72         def update_fun():
---> 73             self.f = fun_wrapped(self.x)
     74
     75         self._update_fun_impl = update_fun

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/_differentiable_functions.py in fun_wrapped(x)
     68         def fun_wrapped(x):
     69             self.nfev += 1
---> 70             return fun(x, *args)
     71
     72         def update_fun():

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/optimize.py in __call__(self, x, *args)
     72     def __call__(self, x, *args):
     73         """ returns the the function value """
---> 74         self._compute_if_needed(x, *args)
     75         return self._value
     76

~/.anaconda/lib/python3.8/site-packages/scipy/optimize/optimize.py in _compute_if_needed(self, x, *args)
     66         if not np.all(x == self.x) or self._value is None or self.jac is None:
     67             self.x = np.asarray(x).copy()
---> 68             fg = self.fun(x, *args)
     69             self.jac = fg[1]
     70             self._value = fg[0]

~/.anaconda/lib/python3.8/site-packages/dipy/align/imaffine.py in distance_and_gradient(self, params)
    783         """
    784         try:
--> 785             self._update_mutual_information(params, True)
    786         except (AffineInversionError, AffineInvalidValuesError):
    787             return np.inf, 0 * self.metric_grad

~/.anaconda/lib/python3.8/site-packages/dipy/align/imaffine.py in _update_mutual_information(self, params, update_gradient)
    670
    671         # Update the histogram with the current joint intensities
--> 672         static_values, moving_values = self._update_histogram()
    673
    674         H = self.histogram  # Shortcut to `self.histogram`

~/.anaconda/lib/python3.8/site-packages/dipy/align/imaffine.py in _update_histogram(self)
    631             pts = sp_to_moving.dot(self.samples.T).T  # Points on moving grid
    632             pts = pts[..., :self.dim]
--> 633             self.moving_vals, inside = self.interp_method(self.moving, pts)
    634             self.moving_vals = np.array(self.moving_vals)
    635             static_values = self.static_vals

dipy/core/interpolation.pyx in dipy.core.interpolation.__pyx_fused_cpdef()

TypeError: No matching signature found
```

oesteban:
Okay, nevermind for now @arokem.

I have moved on to ``antsRegistration`` for the following reasons:
- Easier to read for newbies (thinking of the tutorial)
- Easier to understand for newbies that have already used ANTs or at least know about it (not a dismissable population)
- Easier for me (I use it quite often)
- Easier for debugging (we can focus less on the registration problem)
- Dividing responsibilities (we do not depend fully on DIPY, and maybe make the move with a more stable interface)
- ANTs' finesse - there are a few interesting points in ants 2.3 that I don't think DIPY implements (e.g., you can play with several masks, one per level).

oesteban:
Okay, I think we can get this merged and then see where we go from here (basically, use a non-dummy prediction model)

arokem:
This does significantly change the objectives of the software, which was a pure python implementation of these ideas. 

Regarding API/usage: I think we can sort that out. Making it easier for newbies should be a matter of writing clearer APIs. That should also help you as a developer, I believe.

Regarding finesse: if these features are important for performance, we can implement this in Python. 

Regarding division of responsibility: I am not sure I follow. What's the issue with relying on DIPY?

arokem:
I guess one way we could go about this is to merge this now, and then it would be up to me to refactor this out later on. I'd be OK with that.

arokem:
One issue for the tutorial (and in general) : how do you expect users to install the software? A `pip install` won't work anymore, because you can't pip install ANTS. Can it be conda installed? Or does it have to be docker? Are you assuming ISBI tutorial attendees have docker installed and the permissions to run it on the machine that they are using for the tutorial?

oesteban:
Yes, I agree this raises a whole lot of questions and I also want to see this being python-only.

Conda could be a good trade-off for installation until that happens.

For the tutorial, a docker image should do.

oesteban:
Re: division of responsibilities

If we feel there are some pieces of code that would make sense to upstream into DIPY, it will be easier to do if we focus all the attention (about DIPY) on the models. When that bit is rock-solid, we can then proceed with the registration part. Also, getting good results with ants can be the best way to set a reference quality for DIPY's implementation.

oesteban:
I would like to merge this asap - may I ask for a final round of reviews? @arokem @dPys @sebastientourbier

emc/estimator.py
```diff
@@ -0,0 +1,181 @@
...
+):
+    """
+    Remove outliers at both ends of the intensity distribution and fit into a given dtype.
+
+    This interface tries to emulate ANTs workflows' massaging that truncate images into
+    the 0-255 range, and applies percentiles for clipping images.
+    For image registration, normalizing the intensity into a compact range (e.g., uint8)
+    is generally advised.
+
+    To more robustly determine the clipping thresholds, data are removed of spikes
```
sebastientourbier, line 154: I guess you did mean:
> [...], spikes are removed from data with a median filter.

emc/model.py
```diff
@@ -0,0 +1,216 @@
...
+            self._model.predict(
+                _rasb2dipy(gradient),
+                S0=self._S0,
+                step=step,
+            )
+        )
+        if predicted.ndim == 3:
+            return predicted
+
+        retval = np.zeros_like(self._mask, dtype="float32")
```
sebastientourbier, line 153: It seems here you are making the assumption that `self._mask is not None` as you flatten in this case the data given to the fit, is that correct?

emc/model.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+class TensorModel:
+    """A wrapper of :obj:`dipy.reconst.dti.TensorModel."""
+
+    __slots__ = ("_model", "_S0", "_mask")
+
+    def __init__(self, gtab, S0=None, mask=None, **kwargs):
+        """Instantiate the wrapped tensor model."""
+        from dipy.reconst.dti import TensorModel
```
sebastientourbier, line 104: TensorModel of Dipy and TensorModel of emc.model might be confusing. What about `import dipy.reconst.dti` and then we call ` dipy.reconst.dti.TensorModel`?

emc/dmri.py
```diff
@@ -0,0 +1,218 @@
...
+            The test data/gradient come **from the original dataset**.
+
+        """
+        if not Path(self._filepath).exists():
+            self.to_filename(self._filepath)
+
+        # read original DWI data & b-vector
+        with h5py.File(self._filepath, "r") as in_file:
+            root = in_file["/0"]
+            dwframe = np.asanyarray(root["dataobj"][..., index])
```
dPys, line 75: curious -- why `dwframe` and `bframe` vs. just `test_data` and `test_gradients`?

emc/dmri.py
```diff
@@ -0,0 +1,218 @@
...
+            bframe = np.asanyarray(root["gradients"][..., index])
+
+        # if the size of the mask does not match data, cache is stale
+        mask = np.zeros(len(self), dtype=bool)
+        mask[index] = True
+
+        train_data = self.dataobj[..., ~mask]
+        train_gradients = self.gradients[..., ~mask]
+
```
dPys, line 84: So I think it would still be good to have some mechanism for further filtering which gradient samples are included for model training and testing upon each iteration. On the one hand, there are those samples that are too close in proximity along the sphere to those which we are trying to predict and therefore risk unfavorably biasing the predictions. But even if that's still a matter of debate, there may simply be outlier volumes or gradients that we wish to manually exclude. For that reason alone, I think it would still be beneficial to assume some additional level of control over which samples get used, no?

On that note, I do think that a threshold for filtering out training samples based on q-space proximity should itself be adaptive since I'd assume that the extent of this bias will vary as a function of model choices and sampling scheme (e.g. we should also think about what might happen here in the half-sphere case and how best to accommodate...).

But for now, adding some functionality here that does something like the following might be a good starting point:
```
def train_gradient_filter(
    test_gradients, train_gradients, cutoff=2.0
):
    train_bvec = np.asanyarray(train_gradients[:, 0:3])
    train_bval = np.asanyarray(train_gradients[:, 3])
    test_bvec = np.asanyarray(test_gradients[:, 0:3])
    test_bval = np.asanyarray(test_gradients[:, 3])
    
    min_bval = min(min(train_bval), test_bval)
    all_qvals = np.sqrt(train_bval - min_bval)
    prediction_qval = np.sqrt(test_bval - min_bval)

    # Convert q values to percent of maximum qval
    max_qval = max(max(all_qvals), prediction_qval)
    all_qvals_scaled = all_qvals / max_qval * 100
    scaled_qvecs = train_bvec * all_qvals_scaled[:, np.newaxis]
    scaled_prediction_qvec = test_bvec * (prediction_qval / max_qval * 100)
    
    # Calculate the distance between the training qvecs and the testing qvec
    gradient_mask = (
        np.linalg.norm(scaled_qvecs - scaled_prediction_qvec, axis=1
                       ) > cutoff
    ) * (np.linalg.norm(scaled_qvecs + scaled_prediction_qvec, axis=1
                        ) > cutoff)

    return gradient_mask
    
gradient_mask = train_gradient_filter(test_gradients, train_gradients)
train_data = self.dataobj[..., ~gradient_mask]
train_gradients = self.gradients[..., ~gradient_mask]
```
And then maybe also adding an `exclusion_indices` argument as well for manually dropping training samples if needed.

WDYT @oesteban ?

dPys (changes requested):
Such incredible progress on this! I think users are really going to love the sklearn-style fit, transform, predict methods. I left a few comments and a suggested enhancement, but otherwise just very excited to see this tool maturing, especially with the "Oscar-touch" :-)

emc/estimator.py
```diff
@@ -0,0 +1,52 @@
...
+
+        Parameters
+        ----------
+        X : :obj:`~nibabel.SpatialImage`
+          A DWI image, as a nibabel image object.
+        init : :obj:`~numpy.ndarray`
+          An Nx4x4 array of initialization transforms.
+
+        """
+        loo_index = np.random.shuffle(range(len(X)))
```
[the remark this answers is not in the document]
dPys: Deprecated comment-- see longer response below.

emc/estimator.py
```diff
@@ -0,0 +1,181 @@
...
+            ).to_filename(bmask_img)
+            kwargs["mask"] = dwdata.brainmask
+
+        kwargs["S0"] = _advanced_clip(dwdata.bzero)
+
+        for i_iter in range(1, n_iter + 1):
+            index_order = np.arange(len(dwdata))
+            np.random.shuffle(index_order)
+            with tqdm(total=len(index_order), unit="dwi") as pbar:
+                for i in index_order:
```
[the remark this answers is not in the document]
oesteban: That would parallelize whole orientations. @sebastientourbier is preparing the parallelization of blocks within any given orientation. The idea here is to be able to benefit from previous alignments in the latter alignments.

emc/dmri.py
oesteban: test_data and test_gradients imply there is some cross-validation going on. Because this is just a data representation object, it seemed more adequate not to refer to the framework you are working on.

emc/dmri.py
```diff
@@ -0,0 +1,96 @@
...
+    def to_filename(self, filename):
+        """Write an HDF5 file to disk."""
+        filename = Path(filename)
+        if not filename.endswith(".h5"):
+            filename = filename.parent / f"{filename.name}.h5"
+
+        with h5py.File(filename, "w") as out_file:
+            out_file.attrs["Format"] = "EMC/DWI"
+            out_file.attrs["Version"] = np.uint16(1)
```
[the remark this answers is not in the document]
oesteban: Actually, this file contains the original dataset. I'm still unsure of the I/O because the current implementation allows you to do some funky stuff (e.g., if you change the ``_filepath`` attribute, the behavior is easily problematic because you could potentially store a realigned dataset and that's not what the estimator is expecting).

But yes, happy to grow from here.

emc/model.py
```diff
@@ -0,0 +1,216 @@
...
+
+        elif model.lower().startswith("sfm"):
+            from emc.utils.model import (
+                SFM4HMC as Model,
+                ExponentialIsotropicModel,
+            )
+
+            param = {
+                "isotropic": ExponentialIsotropicModel,
+            }
```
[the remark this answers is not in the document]
oesteban: Let's get this merged and sure, on top of this PR we can definitely do that.

emc/dmri.py
oesteban: > therefore risk unfavorably biasing the predictions.

It is still unclear to me why the bias would be unfavorable in this case

> there may simply be outlier volumes or gradients that we wish to manually exclude.

This is a chicken-and-egg problem. How do you decide what volumes should be excluded? Perhaps they'd be acceptable after head motion? I prefer to make the implementation as agnostic as possible of the rest of the pipeline. This is not to say that we should not look into this issue, but my opinion is that we can only investigate this outlier filtering after we have successfully tried the code on a lot of data. We need to measure how much it improves w.r.t. a baseline. This PR tries to set such a baseline.

Instead of thinking about solutions to hypothetical problems, let's hypothesize about the most relevant problems we expect first, then plan for a test that would evaluate solutions to the problem, and finally once we have both hypothesis and test, make decisions to improve performance.

Finally, and in particular about the code you propose:
- I would not implement this in the DWI object because this is not partitioning the dataset, it is partioning the partition with some additional criteria. IMHO, the responsibility of this falls more on the estimator object side.
- Let's keep this code around and test it when we have a functional prototype (i.e., a baseline as I called it before). This first implementation only looked for setting up the foundations.

oesteban: For instance, one problem I do see coming at us:

Say you have 12 b=1000, 12 b=2000 and 12 b=3000. Eddy currents for the two outer shells will be much larger than those of the b=1000. Because the dataset is already pretty biased towards the outer shells, when aligning one orientation of the b=1000 shell we might actually "overcorrect" for eddy. So in this case, you might want to completely dismiss the largest b (instead of close orientations). WDYT @arokem ?

emc/model.py
oesteban: Yes, I believe it was in my latest commit, I generate a mask by thresholding the data if no mask is given. Therefore, it should not be `None` at this point.

emc/estimator.py
oesteban: (numpy array)

oesteban:
Okay, we will need to push this faster. Happy to discuss all the lines above in future PRs building on top of this one. Thanks for the feedback!

Issue #29, opened 2021-04-10
Uniform names for models


arokem:
I think (based on reading the tutorial in https://github.com/nipreps/nipreps-book/issues/14, that we currently have "Tensor" and "DKI" as model names in the ModelFactory. Might be better to have "DTI", "DKI".

PR #30, opened 2021-04-10
RF: Use consistent names for DTI/DKI.

arokem:
Addresses #29

### DIFF
--- a/eddymotion/model.py
+++ b/eddymotion/model.py
@@ -53,11 +53,8 @@ class ModelFactory:
                 "isotropic": ExponentialIsotropicModel,
             }
 
-        elif model.lower().startswith("dti"):
-            Model = DTIModel
-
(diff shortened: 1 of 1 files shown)


eddymotion/model.py
```diff
@@ -53,8 +53,8 @@ def init(gtab, model="TensorModel", **kwargs):
                 "isotropic": ExponentialIsotropicModel,
             }
 
-        elif model.lower().startswith("tensor"):
-            Model = TensorModel
+        elif model.lower().startswith("dti"):
```
oesteban, line 56: I guess now we can do:
```suggestion
        elif model.lower() == "dti":
```

(same for DKI, btw)

oesteban (approved):
LGTM, just a minimal suggestion.

PR #34, opened 2021-04-12
ENH: Parallelize DTI model fit and prediction

sebastientourbier:
Providing an parallel implementation of tensor model fitting and prediction using `concurrent.futures` and `asyncio` libraries.

It can be executed as the following:
```python
from eddymotion.estimator import EddyMotionEstimator
EddyMotionEstimator.fit(data, model="DTI", n_threads=6)
```


sebastientourbier:
It seems there is a list of commits related to your de-nipype branch and I am not sure exactly what would be correct way to fix it. But github should be able to eliminate them at merge, isn't it? @oesteban

oesteban:
Will look asap - sorry for my slow response

arokem:
Does a rebase on top of main get rid of the extraneous commits?

oesteban:
I'm trying to get the rebase done. Conflicts are a bit tricky. Please hold off from attempting it, @sebastientourbier

arokem (commented):
Looks like users who are calling this with `n_threads` would still take the hit of the overhead associated with the initialization of the `ThreadPoolExecutore`. Is this a big hit? Is there any way of avoiding this without adding a ton of boilerplate?

eddymotion/model.py
```diff
@@ -127,21 +152,86 @@ def __init__(self, gtab, S0=None, mask=None, **kwargs):
...
+                self.fit_chunk,
+                self._model_chunks[i],
+                data_chunks[i]
+            )
+            for i in range(self._n_threads)
+        ]
+
+        try:
+            event_loop = asyncio.get_event_loop()
+            self._model_chunks = event_loop.run_until_complete(asyncio.gather(*fit_tasks))
```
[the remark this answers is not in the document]
oesteban: Indeed, it would be good to narrow down the kinds of errors we might expect here - but, given that there's no `except`, I assume this is to ensure that the `finally` block is executed in all cases.

The exception is then reraised unchanged. Is that it, Seb?

oesteban:
I believe the `ThreadPoolExecutor` does not cause a big overhead. Python documentation generally favors the `ProcessPoolExecutor`, but I do think in this case that would be a bad idea (only for Linux systems - given that `spawn` is not used to fork processes).

eddymotion/model.py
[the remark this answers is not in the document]
sebastientourbier: Good catch! the try/finally is indeed useless here. I got inspired by the example in https://pymotw.com/3/asyncio/executors.html#threads that uses this to ensure that the event loop is closed at the very end. But when called in a jupyter notebook, it was raising ``Error: cannot close a running loop``. More specifically the problem is that `get_event_loop` gets an existing event loop that the call does not have permission to close.

I made a few changes in my code on my side where I am now creating a new event loop. For instance, the code for the fit looks like:
```python
with ThreadPoolExecutor(max_workers=self._n_threads) as executor:
    loop = asyncio.new_event_loop()

    fit_tasks = [
        loop.run_in_executor(
            executor,
            self.fit_chunk,
            self._model_chunks[i],
            data_chunks[i]
        )
        for i in range(self._n_threads)
    ]

    try:
        self._model_chunks = loop.run_until_complete(asyncio.gather(*fit_tasks))
    finally:
        loop.close()
```
In this way we have full control on the event loop and I use the `finally` clause to ensure that the created event loop is closed at the very end.

Alternatively, I could also keep the `get_event_loop` and remove the try/finally clauses and never call the `loop.close()`. WDYT?

I haven't pushed changes yet to github as @oesteban asked me to hold off fixing merge conflicts (https://github.com/nipreps/EddyMotionCorrection/pull/34#issuecomment-819542777)  but happy to push changes corresponding to one of the 2 alternatives when I get green light.

sebastientourbier (commented):
The `ThreadPoolExecutor` does not seem indeed to cause a big overhead and is necessary to handle the threads in parallel with certain concurrency limit (`n_threads`).

Also I think I could make the code more clear with the use of a `with ...:` clause that creates it as follows:
```python
with ThreadPoolExecutor(max_workers=self._n_threads) as executor:
    loop = asyncio.new_event_loop()

    fit_tasks = [
        loop.run_in_executor(
            executor,
            self.fit_chunk,
            self._model_chunks[i],
            data_chunks[i]
        )
        for i in range(self._n_threads)
    ]

    try:
        self._model_chunks = loop.run_until_complete(asyncio.gather(*fit_tasks))
    finally:
        loop.close()
```
WDYT?

oesteban:
Yes sorry for my slow response, but it would indeed be easier if you wait a little longer until I push my merge.

arokem:
It all makes sense to me!

oesteban:
Okay @sebastientourbier, I've rewritten your history (which is a pretty awful thing to do, but the only way to get this straight). Please don't do anything, I'll write instructions to you in the next comment.

oesteban:
So, please follow these steps:
```
git checkout enh/parallel  # do not git pull or anything of the sort!
git checkout -b backup/enh/parallel
git branch -D enh/parallel
git checkout main
git pull
git checkout enh/parallel
```

Now you'll have two branches with all the code. I would suggest you check I did the right merge with:

```
git diff backup/enh/parallel eddymotion/model.py
```

I might have overseen some problems / edits.

eddymotion/estimator.py
```diff
@@ -21,6 +21,7 @@ def fit(
         align_kwargs=None,
         model="b0",
         seed=None,
+        n_threads=1,
```
oesteban, line 24:
```suggestion
        n_threads=None,
```

eddymotion/estimator.py
```diff
@@ -69,6 +72,8 @@ def fit(
 
         kwargs["S0"] = _advanced_clip(dwdata.bzero)
 
+        kwargs["n_threads"] = n_threads
```
oesteban, line 75:
```suggestion
        kwargs["n_threads"] = n_threads or cpu_count()
```

Please add `from os import cpu_count` at the top.

oesteban (commented):
Nitpicks before we confirm my merge is decent. After that, I'll go with a deeper review.

dPys:
Looks great @sebastientourbier ! Can we add a test for this that shows the parallelized version yields the same result as the serial version?

oesteban:
> Looks great @sebastientourbier ! Can we add a test for this that shows the parallelized version yields the same result as the serial version?

Let's say close enough :D

dPys:
> > Looks great @sebastientourbier ! Can we add a test for this that shows the parallelized version yields the same result as the serial version?
> 
> Let's say close enough :D

What Oscar said :-)

setup.cfg
```diff
@@ -23,6 +23,7 @@ python_requires = >=3.7
 install_requires =
     dipy>=1.3.0
     scikit-image>=0.14.2
+    nest-asyncio>=1.5.1
```
oesteban, line 26: do we really need this patch?

sebastientourbier: If not used, I was getting a `RuntimeError: This event loop is already running` when run in a jupyter notebook where the event loop is already running. This patch allows the event loop to be nested.

Here is a little more details extracted from https://pypi.org/project/nest-asyncio/:

> By design `asyncio` does not allow its event loop to be nested. This presents a practical problem: When in an environment where the event loop is already running it’s impossible to run tasks and wait for the result. Trying to do so will give the error “`RuntimeError: This event loop is already running`”.

sebastientourbier:
Thanks @oesteban! Will follow your instructions, double check the merge, and integrate the last changes related to Ariel's comment.
Let you know when I am done!

setup.cfg
sebastientourbier: I made a test with the latest version of my implementation where I am creating a dedicated event loop and it seems it still returning me with the error: `RuntimeError: Cannot run the event loop while another loop is running`

sebastientourbier:
Hi @oesteban @arokem @dPys ! Everything should up-to-date now 😄

oesteban:
@arokem - I've tested sebastientourbier/EddyMotionCorrection#3 and it seems to work on the theoretical POV, but then the call to `TensorModel.fit` (where `TensorModel` is that from DIPY) seems to take the same ~30 sec despite passing in 320k, 80k, or 44k voxels. The `predict` method clearly scales down with the parallelization factor.

I've played with the `step` and it doesn't seem to have any impact. We (I) must be missing something.

Issue #41, opened 2021-04-14
Implement the ``AverageDWModel`` we propose in the nipreps-book


oesteban:
This model could easily compete with the b=0 model.

Further sophistication to the current naive implementation in the book:
- [ ] Average only volumes with b < threshold.
- [ ] Account for signal drift and b-value before averaging
- [ ] Average with robust statistics (median?)

PR #50, opened 2021-04-26
enh : Add regression to the mean model

celprov:
Naive implementation of the AverageDWModel proposed in the nipreps-book. #41

### DIFF
--- a/eddymotion/model.py
+++ b/eddymotion/model.py
@@ -90,6 +90,22 @@ class TrivialB0Model:
         """Return the *b=0* map."""
         return self._S0
 
+class AverageDWModel:
+    """A trivial model that returns an average map."""
+
+    __slots__ = ("_data",)
+
+    def __init__(self, gtab, **kwargs):
+        """Implement object initialization."""
+        return  # do nothing at initialization time
+
+    def fit(self, data, **kwargs):
+        """Calculate the average."""
+        self._data =  data.mean(-1)
+
+    def predict(self, gradient, **kwargs):
+        """Return the average map."""
+        return self._data
 
 class DTIModel:
     """A wrapper of :obj:`dipy.reconst.dti.TensorModel."""
--- a/eddymotion/model.py
+++ b/eddymotion/model.py
@@ -94,15 +94,15 @@ class TrivialB0Model:
 class AverageDWModel:
     """A trivial model that returns an average map."""
 
-    __slots__ = ("_data",)
+    __slots__ = ("_data", "_gtab")
 
     def __init__(self, gtab, **kwargs):
         """Implement object initialization."""
-        return  # do nothing at initialization time
+        self._gtab = gtab
 
     def fit(self, data, **kwargs):
         """Calculate the average."""
-        self._data = data.mean(-1)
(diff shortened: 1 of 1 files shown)


arokem (commented):
One small comment from me: the name `AverageDWModel` suggests that this is a model that averages the diffusion-weighted signals, but in practice, this averages both diffusion-weighted *and* non-diffusion-weighted (i.e., b0) signals.

One related suggestion is to give this model a bit more flexibility by allowing some arguments that will control which b-values are used in averaging (for example, `lb` and `ub`  keyword arguments that specify a range)

oesteban:
> One small comment from me: the name `AverageDWModel` suggests that this is a model that averages the diffusion-weighted signals, but in practice, this averages both diffusion-weighted _and_ non-diffusion-weighted (i.e., b0) signals.

In principle, this should only receive DW volumes (i.e., no b=0). However, I admit that I created a little loophole for it here: https://github.com/nipreps/eddymotion/blob/c7de9f7c4d687c7064bc321fad829479b0730af6/eddymotion/dmri.py#L50

Effectively, if you generated the training set with `with_b0=True`, you'd have a b=0 at the beginning of the dataset.

Because checking the gradient table to see if there are or aren't b=0 volumes is almost free, I'll make a quick suggestion for that.

> One related suggestion is to give this model a bit more flexibility by allowing some arguments that will control which b-values are used in averaging (for example, `lb` and `ub` keyword arguments that specify a range)

Yes, I had already discussed this with @celprov, and I warned her that I would give this a thorough review including suggestions (including this one among others). The idea was to create a base of a PR as an exercise and build from there, given that she is starting with GitHub, coding, etc.

eddymotion/model.py
```diff
@@ -91,6 +91,24 @@ def predict(self, gradient, **kwargs):
         return self._S0
 
 
+class AverageDWModel:
+    """A trivial model that returns an average map."""
+
+    __slots__ = ("_data",)
```
oesteban, line 97:
```suggestion
    __slots__ = ("_data", "_gtab")
```

eddymotion/model.py
```diff
@@ -91,6 +91,24 @@ def predict(self, gradient, **kwargs):
...
+
+    def __init__(self, gtab, **kwargs):
+        """Implement object initialization."""
+        return  # do nothing at initialization time
```
oesteban, line 101:
```suggestion
        self._gtab = gtab
```

eddymotion/model.py
```diff
@@ -91,6 +91,24 @@ def predict(self, gradient, **kwargs):
...
+
+    def fit(self, data, **kwargs):
+        """Calculate the average."""
+        self._data = data.mean(-1)
```
oesteban, line 105:
```suggestion
        self._data = data[..., self._gtab[..., 3] > 50].mean(-1)
```

eddymotion/model.py
oesteban: keep a reference to the gradient table, so that later you can check on it when creating the average.

oesteban (approved):
Looking good! Added some suggestions addressing @arokem's concern. Thanks much for this @celprov !

Issue #40, opened 2021-04-14
Integration and validation tests with b=0 only


oesteban:
Two low-hanging fruit tests using only a b=0:

- Validation:
    - [x] Unit tests of the `TrivialB0Model` implementation
    - [x] Unit tests of ANTs' config for *b=0*.
- Integration:
    - [x] Introduce a series of transforms (e.g., copy them from another dataset) and resample the *b=0* through that motion series. Then run the whole estimator and check the proximity of the estimated transforms.

oesteban:
Okay, I've tried to set up some basics, and implemented a unit-test for the `TrivialB0Model` sketching what we intend to do here:

https://github.com/nipreps/eddymotion/blob/152e0d4aeb6917965ed1821ea1507fb68630fa3b/eddymotion/tests/test_model.py#L29-L42

> Unit tests of ANTs' config for b=0.

I see this implemented in [`test_estimator.py`](https://github.com/nipreps/eddymotion/blob/main/eddymotion/tests/test_estimator.py). To do so:

- [ ] Upload an example B0 map (3D)
- [ ] Apply random (and small) rotations and translations on the image. These can be generated with `nibabel.eulerangles.euler2mat` (e.g., https://github.com/poldracklab/nitransforms/blob/master/nitransforms/tests/test_linear.py#L107). For the resampling, I would use `nitransforms.linear.Affine.resample()`.
- [ ] Run the registration interface loading the parameters from config
- [ ] Check that the error of the affine is below some threshold when transforming points scattered across the volume.

>  Introduce a series of transforms (e.g., copy them from another dataset) and resample the b=0 through that motion series. Then run the whole estimator and check the proximity of the estimated transforms.

I see this implemented in [`test_integration.py`](https://github.com/nipreps/eddymotion/blob/main/eddymotion/tests/test_integration.py). To do so:
- [ ] Upload an example B0 map (3D) - or reuse from the previous test
- [ ] Generate a series of moved B0's with the realignment matrices from another, real dataset.
- [ ] Initiate the `TrivialB0Model` with this B0 map and feed the dwi object with the moved data (simulate perhaps several transforms).

oesteban:
Hi there! I've uploaded some data under https://github.com/nipreps/eddymotion/tree/main/eddymotion/tests/data, generated by @dPys with FiberFox.

sebastientourbier:
@oesteban I am progressing! For now, I managed to:
[...]

PR #59, opened 2021-11-11
ENH: Add integration test with data for EddyMotionEstimator using the trivial b0 model

sebastientourbier:
Part of #40, this PR includes the following:

*  A notebook in `docs/notebooks` that I created to reproduce all the steps required for the creation of the data used in the integration. All details for the generation can be found there.

* A set of files for the integration test:
  * `b0.moving.nii.gz`: 4D volume where the first 3D volume is the original b0, followed by 10 b0 volumes transformed with the motion parameters estimated from the last 10 frames of the fMRI recording
  * `gradients.moving.tsv`: Fake RAS+B gradient table for the 10 moving b0 volumes
  * `b0.moving.transforms.npy`: A Numpy array of size (4x4x10) that stores the transforms estimated from fMRI
 
* A new function `test_proximity_estimator_trivial_model()` that uses the set of files and checks the proximity of the transforms estimated by `EddyMotionEstimator` with a trivial B0 model.


eddymotion/tests/test_integration.py
```diff
@@ -21,3 +21,51 @@
...
+
+    # For each moved b0 volume
+    for i, xfm in enumerate(em_affines):
+        fixed_b0_img = nb.Nifti1Image(_b0, affine=_affine)
+        xfm2 = nit.linear.Affine(
+            ref_xfms[..., i],
+            reference=fixed_b0_img
+        )
+        assert np.all(
+            abs(xfm.map(xfm.reference.ndcoords.T) - xfm2.map(xfm.reference.ndcoords.T)) < 0.4
```
dPys, line 70: Where is 0.4 coming from out of curiosity? I'm guessing this is just unique to this test-data case?

dPys (approved):

Issue #53, opened 2021-09-14
Implement Gaussian Process as a model


oesteban:
In order to compare baseline performance vs. FSL eddy.

josephmje:
Adding notes from our 2021-12-07 meeting here:

A first stab at implementing the Gaussian process model is at #60. 

At the moment, we haven't done any tuning or optimization of the model. If we are going to eventually compare it to FSL eddy, that will need to be done.

For testing, some suggestions are to use:
- simulated data from FiberFox
- [IXI dataset](https://osf.io/8k95s)
- a high quality dataset with certain voxels masked (perhaps near a white matter tract or some CSF in the middle of a ventricle)

Will eventually need to add better parallelization. Currently, the DTI model chunks the data before fitting the model. We can also implement that code here. For the future, we may need to dig into dipy and can use joblib/dask to further improve the parallelization.

dPys:
@josephmje - Not sure if you or anyone else are still working on the GP model, but figured I'd chime in with a few thoughts to get the conversation going again:

The GP model needs at least a PoC design for the kernel before it can be tuned. Here's a first template for what that might look like. High-Level Requirements:
*Isotropism, uniform affinity for all angular directions.
*Symmetry around the origin.

The way that most folks have seemed to address this in the context of q-space is to factorize the covariance function into (a) an angular part and (b) a radial part, with the addition of (c) an error term. 

The angular part can be constructed using Legendre polynomials, whereby symmetry about the origin can be achieved by excluding odd terms and ensuring that their coefficients are positive. Past work seems to have used an order <=6, with the consensus being that grabbing the first four terms is best.

Maybe roughly something along the lines of:
```
from numpy.polynomial.legendre import Legendre

def angular_polynomials(n_gradients, order=6):
    """
    A real continuous function $C(\theta)$ is a valid covariance function on the 
    sphere if and only if it is of the form: 
    $\sum_{n=0}^{\infty} a_{n}P_{n}(\cos(\theta))$ 
    where $\theta \in (0, \pi)$
    """
    theta = np.arange(n_gradients)/n_gradients * np.pi
    num_pol = range(0, order + 1, 2)
    y = np.ones((len(num_pol),len(theta)))   
    coeff = np.eye(int(order/2 + 1))
    
    for i, n in enumerate(num_pol):
        leg = Legendre(np.abs(coeff[i]))
        y[i] = leg(np.cos(theta))
    return y
```

[...]

Issue #37, opened 2021-04-14
Port SFM model back from the legacy branch


oesteban:

josephmje:
Isotropic fit can be used for single-shell data but for multi-shell data, will need to use exponential isotropic fit. Dipy should have some code for detecting whether data is single or multi-shell data. @arokem mentioned it's being used in the DKI model to test whether the data is eligible.

arokem:
This is the function that checks if data are multi-b: https://github.com/dipy/dipy/blob/be956a529465b28085f8fc435a756947ddee1c89/dipy/core/gradients.py#L825

josephmje:
Reporting some weirdness with the SFM model currently:

If including a mask, run into memory issues, even on the small IXI dataset:
```
MemoryError: Unable to allocate 735. GiB for an array with shape (128, 128, 56, 128, 56, 15) and data type float64
```

If no mask, run into some looping issues.
```
AttributeError                            Traceback (most recent call last)
<ipython-input-15-c2464056eea0> in <module>
----> 1 em_affines = EddyMotionEstimator.fit(dmri_dataset, model="SFM", omp_nthreads=16)

/mnt/tigrlab/scratch/mjoseph/eddymotion/eddymotion/estimator.py in fit(dwdata, n_iter, align_kwargs, model, omp_nthreads, seed, **kwargs)
     97 
     98                         # fit the model
---> 99                         dwmodel.fit(data_train[0])
    100 
    101                         # generate a synthetic dw volume for the test gradient

/mnt/tigrlab/scratch/mjoseph/eddymotion/eddymotion/model.py in fit(self, data, **kwargs)
     90 
     91             try:
---> 92                 self._model = loop.run_until_complete(
     93                     asyncio.gather(*fit_tasks))
     94             finally:

/scratch/mjoseph/.pyenv/versions/3.8.1/envs/ohbm_venv/lib/python3.8/site-packages/nest_asyncio.py in run_until_complete(self, future)
     79                 raise RuntimeError(
     80                     'Event loop stopped before Future completed.')
---> 81             return f.result()
     82 
     83     def _run_once(self):

/scratch/mjoseph/.pyenv/versions/3.8.1/lib/python3.8/concurrent/futures/thread.py in run(self)
     55 
     56         try:
---> 57             result = self.fn(*self.args, **self.kwargs)
     58         except BaseException as exc:
     59             self.future.set_exception(exc)

[...]

PR #60, opened 2021-12-02
ENH: Adds Sparse Fascicle and Gaussian Process models

josephmje:
Addresses #37 and #53


eddymotion/model.py
```diff
@@ -53,15 +53,17 @@ def init(gtab, model="DTI", **kwargs):
...
+                    "solver": "ElasticNet",
+                    "isotropic": ExponentialIsotropicModel,
+                }
+            else:
+                from sklearn.gaussian_process import GaussianProcessRegressor
 
-            param = {
-                "isotropic": ExponentialIsotropicModel,
-            }
+                param = {"solver": GaussianProcessRegressor}
```
dPys, line 66: In the long-term, I'd bet we get more miles out of using https://docs.pymc.io/en/v3/Gaussian_Processes.html in place of sklearn's GaussianProcessRegressor but for now this seems fine!

RE: tuning this, we could start by trying to make alpha very small like what we did with the elastic-net case, but I don't know about the kernel hyperparameters. @arokem -- would the idea here be that by wrapping this in SFM, we'd avoid the need for learning the shape of a manual kernel? The default used for GP in sklearn is `ConstantKernel(1.0, constant_value_bounds="fixed" * RBF(1.0, length_scale_bounds="fixed")`.

If we're sticking with this, then I think that either way, we should remove the `fixed` option since "kernel hyperparameters are optimized during fitting unless the bounds are marked as `fixed`" and from Andersson & Sotiropoulos 2015: "The smoothness is determined from hyperparameters whose values are determined directly from the data" and "the hyperparameters will be such that the predictions are a smooth function of gradient direction and, in the case of multi-shell data, b-value." Also noteworthy was this point: "the data given by original hyperparameters is very sharp and even for a dataset with 300 points will give almost half the weight to the center-point (i.e., when predicting the signal for a diffusion direction half the information comes from the point itself)."

dPys: And for selecting the kernel, https://www.sciencedirect.com/science/article/pii/S1053811915006874?via%3Dihub#:~:text=Finding%20k(x,scaling%20or%20variance. they suggest using either exponential or spherical and then optimizing with maximum likelihood

eddymotion/model.py
```diff
@@ -332,6 +334,59 @@ def predict(self, gradient, **kwargs):
...
+        self._solver = solver
+        if solver is None:
+            self._solver = "ElasticNet"
+
+        kwargs = {k: v for k, v in kwargs.items() if k in ("solver",)}
+        self._model = SparseFascicleModel(gtab, **kwargs)
+
+    def fit(self, data, **kwargs):
+        """Clean-up permitted args and kwargs, and call model's fit."""
+        self._model = self._model.fit(data[self._mask, ...])
```
arokem, line 372: For parallelization, we can divide the data up into chunks before calling this, and then call this separately on each chunk (on a separate thread/process)

docs/notebooks/Testing GP model.ipynb
```diff
@@ -0,0 +1,352 @@
...
+   "execution_count": 2,
+   "id": "750f9a01-aac3-452e-8fec-0fcc4034d448",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "base_dir = Path(\"/Users/michael/projects/datasets/ds000206\")\n",
+    "bids_dir = base_dir / \"bids\"\n",
+    "derivatives_dir = base_dir / \"dmriprep\"\n",
+    "\n",
+    "dwi_file = bids_dir / \"sub-05\" / \"ses-JHU1\" / \"dwi\" / \"sub-05_ses-JHU1_acq-GD72_dwi.nii.gz\"\n",
```
oesteban, line 44: Does this file contain *b=0* volumes? I believe it should not (i.e., our dwi object doesn't filter them out)

eddymotion/model.py
```diff
@@ -53,15 +53,17 @@ def init(gtab, model="DTI", **kwargs):
...
-                SFM4HMC as Model,
-                ExponentialIsotropicModel,
-            )
+        elif model.lower() in ("sfm", "gp"):
+            Model = SparseFascicleModel
+            if model.lower() == "sfm":
+                param = {
+                    "solver": "ElasticNet",
+                    "isotropic": ExponentialIsotropicModel,
+                }
```
oesteban, line 62:
```suggestion
            param = {
                "solver": "ElasticNet",
                "isotropic": ExponentialIsotropicModel,
            }
```

eddymotion/model.py
```diff
@@ -53,15 +53,17 @@ def init(gtab, model="DTI", **kwargs):
...
+            else:
```
oesteban, line 63:
```suggestion
            if model.lower() == "gp":
```

oesteban (approved):

docs/notebooks/Testing GP model.ipynb
josephmje: Yes it does. I think the load function filters out b values less than 50. 

https://github.com/nipreps/eddymotion/blob/main/eddymotion/dmri.py#L224-L226

eddymotion/model.py
josephmje: Thanks @arokem, I'm going to hand it over to @dpys to work on the chunking in a separate PR

docs/notebooks/Testing GP model.ipynb
josephmje: We just realized that the load function as it currently stands affects the AverageB0 model since no b0s will be found. Will address in a separate PR.

dPys (approved):
LGTM!

docs/notebooks/Testing GP model.ipynb
oesteban: Is there an AverageB0 model? I think it's AverageDWI

oesteban: Meaning, it actually doesn't expect bzeros

eddymotion/model.py
dPys: @arokem and @oesteban -- Did we ever decide on whether the best approach for this would be to migrate @sebastientourbier 's asyncio parallelization routines for the DTI model to a base class for parallel fit/predict that could be imported to each of the SFM and DKI API's as well? 

On a separate note-- would it be better to stick with asyncio, switch to joblib, or expose both approaches as optional?

oesteban:
I think we should reuse Seb's solution

eddymotion/model.py
arokem: I would stick with a copy of Sebastien's solution here. Let's worry about a more elegant OO design after we see that this works well here.

PR #67, opened 2022-06-03
Restructuring the package

### LINKED ISSUE #63: Restructuring this package to fit newer python packaging standard
I think it may be valuable to restructure this package to [a newer standard](https://packaging.python.org/en/latest/tutorials/packaging-projects/).
It should be relatively easy to do and will establish a "cleaner" structure.

Using [some cookie-cutter template](https://github.com/ionelmc/cookiecutter-pylibrary) may do the trick, but I would love to get your input on this suggestion before I start working on it :satisfied:

Thanks in advance!

GalBenZvi:
This includes mainly the following changes:

* Staging the source code under src/ rather than the package's base.
* Tests are also located there (in proximity to the source code, as suggested by @oesteban)
* Setup configurations are mostly the same as the original package.
* I've added [Github Action](https://github.com/GalBenZvi/eddymotion/blob/restructure/.github/workflows/github-actions.yml) and [pre-commit hooks](https://github.com/GalBenZvi/eddymotion/blob/restructure/.pre-commit-config.yaml) so that it'll be easier for you to control incoming suggestions, especially in light of the coming Hackathon at the OHBM.
* Added sphinx-appropriate documentation structure (docs folder and its derivatives).

I of course made sure that no original functionality was damaged due to these suggestions. Most tests are passing and I think those that do not also fail for the original structure (i.e, it's not a result of the restructuring).

Fixes #63.


effigies:
Just some quick thoguhts from looking at the remote branch (I agree this PR is hard to review as is):

1) We've gone from `eddymotion/` to `src/eddymotion` without anything else in `src`. I don't see any value in this change unless it allows us to separate out extra files that are related to the repository versus related to the package, which this PR does not do.
2) I value having tests in the repository as first-class code that can be run after installing via a wheel. I understand this is not universal in the Python community, but it is highly consistent in the nipy ecosystem.
3) Extra files:
   * .coveragerc and pytest.ini can be replaced with sections in setup.cfg (there actually already is pytest config in there), possibly `pyproject.toml` at this point.
   * .cookiecutterrc, .bumpversion.cfg shouldn't be here. I don't see a value to re-initializing a cookie cutter, and we don't use bumpversion. This project appears to be using `setuptools_scm`.
   * I *am* pro .pre-commit-config.yaml as a helper to maintain styles, but I think it should be part of its own PR where the style is consistently applied *once*. Otherwise you end up with unrelated style changes infecting otherwise reasonably scoped PRs.

GalBenZvi:
Thank you for your notes @oesteban and @effigies.
I'll break this PR into a few smaller ones (and try to include only changes that make sense :sweat_smile:)

Just replying to @effigies regarding the new structure:
> 1. We've gone from `eddymotion/` to `src/eddymotion` without anything else in `src`. I don't see any value in this change unless it allows us to separate out extra files that are related to the repository versus related to the package, which this PR does not do.

It's true that currently nothing else is staged under `src/`, but I think that as this package evolves, it will be useful to break it down to its different modules (for example, motion correction for images originating from different sources such as PET, fMRI or dMRI, as discussed yesterday with @oesteban).
I think locating these modules under different bases in `src/` may be helpful in terms easily navigating this package.

> 2. I value having tests in the repository as first-class code that can be run after installing via a wheel. I understand this is not universal in the Python community, but it is highly consistent in the nipy ecosystem.

I agree - I kept the original relative location of tests dealing with functionality of specific modules, and added a base for tests dealing with functionality of the entire package (like CLI and other features that may be added over time).

> 3. Extra files:
>    
>    * .coveragerc and pytest.ini can be replaced with sections in setup.cfg (there actually already is pytest config in there), possibly `pyproject.toml` at this point.
>    * .cookiecutterrc, .bumpversion.cfg shouldn't be here. I don't see a value to re-initializing a cookie cutter, and we don't use bumpversion. This project appears to be using `setuptools_scm`.
>    * I _am_ pro .pre-commit-config.yaml as a helper to maintain styles, but I think it should be part of its own PR where the style is consistently applied _once_. Otherwise you end up with unrelated style changes infecting otherwise reasonably scoped PRs.

Yes, my bad :sweat_smile:. I'll keep these irrelevant files out of future PR, and make a separate one for pre-commit configurations (and possibly Github Actions?)

I will obviously work under any standard you feel most comfortable with, so let me know which of these suggestions (if any) are welcome (:.

effigies:
> It's true that currently nothing else is staged under src/, but I think that as this package evolves, it will be useful to break it down to its different modules (for example, motion correction for images originating from different sources such as PET, fMRI or dMRI, as discussed yesterday with @oesteban). I think locating these modules under different bases in src/ may be helpful in terms easily navigating this package.

So the idea is to build multiple Python packages from a single source directory? That would make some sense, but I don't think I've encountered machinery to use a single setup.py (or other build config) to do that. Is there an example of a repository doing what you're proposing?

GalBenZvi:
That wasn't quite what I meant. I guess I thought this `src/` structure was an almost-accepted standard since I saw it in some packages (like [dipy](https://github.com/dipy/dipy)) and found it pretty clear to use.
However, I now see that many (probably most) other packages haven't embraced this structure, so I'll leave It to you to decide what's your structure of choice :laughing:.

Sorry for the misunderstanding.

effigies:
Thanks. It looks like dipy is doing the same as us for Python but has some C in src/.

matthew-brett:
One other nice advantage of the `src/` structure is that it allows you to safely import the module from the repository root directory, without accidentally picking up the unbuilt source.  There's some discussion [here](https://hynek.me/articles/testing-packaging).  I must admit, I had been considering switching to that as a default for my own packages.

PR #75, opened 2022-06-24
A small patch for the rebasing of the main directory under src

### LINKED ISSUE #63: Restructuring this package to fit newer python packaging standard

GalBenZvi:
Just a small fix - updated the location of the `._version.py` written by `setup_scm`. Fixes #63


arokem:
Test data now lives here: https://zenodo.org/record/6721110#.YrXPyRPMLKp

arokem:
The tests on this branch are still pretty hopelessly broken. I tried to work towards fixing them in https://github.com/GalBenZvi/eddymotion/pull/3, but there are still failures I don't understand. I would need @oesteban to explain to me how the logo implementation works before I can make progress. In particular, how does the model `fit` method know that only some of the data is provided as training data (e.g., [on this line](https://github.com/nipreps/eddymotion/blob/main/src/eddymotion/tests/test_model.py#L107)) and to account for that fact when selecting which volumes conform with b0 requirements that are set (for example) on [this line](https://github.com/nipreps/eddymotion/blob/main/src/eddymotion/model.py#L276)?

.gitignore
```diff
@@ -113,3 +113,6 @@ ENV/
 
 #kubernetes stuff
 kubernetes/jobs/
+
+# Version files
+*_version.py
```
oesteban, line 118: Better be specific, right?
```suggestion
src/eddymotion/_version.py
```

oesteban (approved):

oesteban:
@arokem It seems we should work on the documentation actually - if you can't wrap your head around how it works then I'm pretty positive that literally any other outsider will fall into that trap. I'm opening #76 for this.

oesteban:
Check out #84. In particular, there was a bug using the gradient table of the full dataset instead of that of the training subsample (i.e., N - 1).

arokem:
Thanks! That helps! See questions on #84 for followup.

PR #79, opened 2022-06-29
FIX: Test structure / pytest & fixtures

oesteban:
Now, at least, tests are run seemingly correctly.
We though still need to setup something to pull data down to make it available to tests.

Continues fixing from #63


arokem:
A couple of general questions: Given non-python dependencies (i.e., ANTS), how do you run the tests locally on your machine? I guess you need to build a docker image and run the tests in the image? Should we add a docker configuration to the repository (we don't have one yet, right?)? I guess the docker requirement would also make running tests on GHA complicated? Maybe we add a Makefile with rules for these kinds of things?

oesteban:
> how do you run the tests locally on your machine?

I have ANTS installed (:

> I guess you need to build a docker image and run the tests in the image? Should we add a docker configuration to the repository (we don't have one yet, right?)?

This shouldn't harm - we do have a base miniconda image for nipreps that would help much here.

> I guess the docker requirement would also make running tests on GHA complicated?

With only ANTS I would just install it in one step of the workflow. That said, GHA with containers is pretty easy too.



> Maybe we add a Makefile with rules for these kinds of things?

That would be a best practice. Esp. now with PEP518, that would clarify how we are expected to build the package (otherwise, you would need to go to the pythonpackage.yml GHA I have recently added and figure yourself).

arokem:
I found out that I can install ANTS using conda: https://anaconda.org/aramislab/ants, so I did that and now the tests run. 

I do still get some test failures that I don't entirely understand. In particular in `test_two_initialisations`. 

I've done some work in https://github.com/GalBenZvi/eddymotion/pull/3 to try to deal with some of these failures. I am not sure whether I'm heading in the right direction, though.

oesteban:
> I found out that I can install ANTS using conda: https://anaconda.org/aramislab/ants, so I did that and now the tests run.

Ha! Didn't know this package existed - thanks for sharing.



> I do still get some test failures that I don't entirely understand. In particular in `test_two_initialisations`.

I think I have fixed something on that test. Now it should only need the data to be accessible.

> I've done some work in https://github.com/GalBenZvi/eddymotion/pull/3 to try to deal with some of these failures. I am not sure whether I'm heading in the right direction, though.

I would suggest rebasing to current main so that we are certain of what is going on.

PR #84, opened 2022-07-01
FIX: Miscelaneous revisions to make tests execute properly

oesteban:
Although this is not going to make all tests to pass, it at least ensures tests can be executed provided with test data fount under ``$TEST_DATA_HOME``.

This PR also fixes a bug on the ``AverageDWModel`` that resulting in the indexing error.

### DIFF
--- a/src/eddymotion/model.py
+++ b/src/eddymotion/model.py
@@ -240,7 +240,7 @@ class TrivialB0Model:
 class AverageDWModel:
     """A trivial model that returns an average map."""
 
-    __slots__ = ("_data", "_gtab", "_th_low", "_th_high", "_bias", "_stat")
+    __slots__ = ("_data", "_th_low", "_th_high", "_bias", "_stat")
 
     def __init__(self, gtab, **kwargs):
         r"""
@@ -264,18 +264,19 @@ class AverageDWModel:
             Whether the summary statistic to apply is ``"mean"`` or ``"median"``.
 
         """
-        self._gtab = gtab
         self._th_low = kwargs.get("th_low", 50)
-        self._th_high = kwargs.get("th_high", self._gtab[3, ...].max())
+        self._th_high = kwargs.get("th_high", 10000)
         self._bias = kwargs.get("bias", True)
         self._stat = kwargs.get("stat", "median")
 
     def fit(self, data, **kwargs):
         """Calculate the average."""
+        gtab = kwargs.pop("gtab", None)
         # Select the interval of b-values for which DWIs will be averaged
-        b_mask = (self._gtab[3, ...] >= self._th_low) & (
-            self._gtab[3, ...] <= self._th_high
-        )
+        b_mask = (
+            (gtab[3] >= self._th_low)
+            & (gtab[3] <= self._th_high)
+        ) if gtab is not None else np.ones((data.shape[-1], ), dtype=bool)
         shells = data[..., b_mask]
 
         # Regress out global signal differences
--- a/test/test_model.py
+++ b/test/test_model.py
@@ -47,50 +47,51 @@ def test_trivial_model():
 def test_average_model():
     """Check the implementation of the average DW model."""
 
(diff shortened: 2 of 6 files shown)


src/eddymotion/model.py
```diff
@@ -240,7 +240,7 @@ def predict(self, gradient, **kwargs):
 class AverageDWModel:
     """A trivial model that returns an average map."""
 
-    __slots__ = ("_data", "_gtab", "_th_low", "_th_high", "_bias", "_stat")
+    __slots__ = ("_data", "_th_low", "_th_high", "_bias", "_stat")
 
     def __init__(self, gtab, **kwargs):
```
arokem, line 245: Do you want to also remove the `gtab` input here?

```suggestion
    def __init__(self, **kwargs):
```

arokem: If yes, don't forget to remove it in the docstring as well.

src/eddymotion/model.py
```diff
@@ -264,18 +264,19 @@ def __init__(self, gtab, **kwargs):
             Whether the summary statistic to apply is ``"mean"`` or ``"median"``.
 
         """
-        self._gtab = gtab
         self._th_low = kwargs.get("th_low", 50)
-        self._th_high = kwargs.get("th_high", self._gtab[3, ...].max())
+        self._th_high = kwargs.get("th_high", 10000)
```
arokem, line 268: I know this is just a toy model, but we might want to document the default values for these, in case they get inherited into other places.

test/test_model.py
```diff
@@ -47,50 +47,51 @@ def test_trivial_model():
...
 
     tmodel_mean = model.AverageDWModel(gtab=gtab, bias=False, stat="mean")
     tmodel_median = model.AverageDWModel(gtab=gtab, bias=False, stat="median")
     tmodel_1000 = model.AverageDWModel(
-        gtab=gtab, bias=False, th_high=1000, th_low=1000
+        gtab=gtab, bias=False, th_high=1000, th_low=900
     )
     tmodel_2000 = model.AverageDWModel(
-        gtab=gtab, bias=False, th_high=2000, th_low=1000
+        gtab=gtab, bias=False, th_high=2000, th_low=900, stat="mean",
```
arokem, line 71: Ha! I just noticed that the default statistic for the "mean" model is the median. Isn't that a bit confusing?

test/test_model.py
```diff
@@ -47,50 +47,51 @@ def test_trivial_model():
...
-    tmodel_1000.fit()
-    tmodel_2000.fit()
+    tmodel_median.fit(data[..., 1:], gtab=gtab[1:].T)
+    tmodel_1000.fit(data[..., 1:], gtab=gtab[1:].T)
+    tmodel_2000.fit(data[..., 1:], gtab=gtab[1:].T)
 
     # Verify that the right statistics is applied and that the model discard b-values < 50
-    assert np.all(tmodel_mean.predict() == np.mean(gtab_w25[:, :2], axis=0))
-    assert np.all(tmodel_median.predict() == np.median(gtab_w25[:, :2], axis=0))
+    assert np.all(tmodel_mean.predict([0, 0, 0]) == 950)
```
arokem, line 82: Isn't the input to `predict` supposed to be a gradient table? That is, a 4-by-n array? Why does it accept a 3-vector?

arokem (commented):
Great. I think that I understand this better now. In general, it seems that these model `fit` and `predict` methods should always take gradient table inputs (i.e. 4-by-n arrays). I note a place where this is not the case, and I wonder whether it should be enforced more strictly.

arokem:
I will go ahead and merge this, so that we can keep working on this on our end with functioning tests. We can come back and fix my comments on a separate PR.

oesteban:
Thanks, I'm coming back to this tomorrow and finishing up whatever is still missing. Running tests locally right now after merge.

PR #85, opened 2022-08-25
FIX: Integration test with B0 (trivial) model

oesteban:
Drops the use of a real AFNI series of transforms. Instead, generate a number of large-yet-plausible transformations for head motion.

I have visually checked a few things:
- NiTransforms correctly generates the dataset with motion, and when *realigining* it with the originally created transforms, indeed the dataset does not show any motion (uncomment line writing out `ground_truth.nii.gz` to see). Meaning, NiTransforms seems to be a reliable tool for this test.
- The correction by our estimator is ATM not great, but we are just testing the integration of all pieces (see #40 for comments on fine-tuning).

Follow-up: #84.
Part-of: #40.


oesteban:
Merging this makes all tests pass locally :)

Issue #86, opened 2022-08-25
Bad initialization of temporal filesystem support for H5 objects


oesteban:
Using `attrs` initialization and setting the default value with mkdtemp does not generate a new name for every new instance. Instead, the first default value is generated and kept when creating all later instances.

Catched this problem when independent tests loading the same original h5 file crashed only if executed sequentially.

PR #87, opened 2022-08-25
FIX: Generate temporary H5 cache path anew with every new instance of ``DWI``

oesteban:
This has been a sneaky one.

When directly generating the default value, ``mkdtemp()`` was executed only once, so all ``DWI`` objects created within the same scope pointed at the same path.

Resolves: #86.


oesteban:
I'm extremely confident this patch needs to go in. Opened the PR just as a heads-up of this sneaky b***d. Merging...

Issue #89, opened 2022-08-26
Replace nipype ants interfaces with antspy?


arokem:
I've recently had the pleasure of working with [antspy](https://antspy.readthedocs.io/en/latest/). It's fast and the API is both straightforward and flexible. I wonder whether we should use this directly in the estimator module, instead of the nipype interface. Benefits include the fact that it's a pip install, so we would remove a pretty heavy dependency (and for many users, the docker dependency!).

arokem:
This is a good place to start: https://github.com/nipreps/eddymotion/blob/main/src/eddymotion/estimator.py#L119

PR #90, opened 2022-09-26
Replace registration to use ANTsPy instead of Nipype

teresamg:
Addresses #89

### DIFF
--- a/src/eddymotion/estimator.py
+++ b/src/eddymotion/estimator.py
@@ -5,13 +5,14 @@ from tempfile import TemporaryDirectory, mkstemp
 import nibabel as nb
 import nitransforms as nt
 import numpy as np
+from nipype.interfaces.ants.registration import Registration
+from pkg_resources import resource_filename as pkg_fn
 from tqdm import tqdm
 
 from eddymotion.model import ModelFactory
-import ants
-import ants.utils.process_args as ants_utils
-import json
-from eddymotion.utils import antsimage_from_path
+
+import pickle
+import sys
 
 
 class EddyMotionEstimator:
@@ -30,7 +31,6 @@ class EddyMotionEstimator:
     ):
         r"""
         Estimate head-motion and Eddy currents.
-
         Parameters
         ----------
         dwdata : :obj:`~eddymotion.dmri.DWI`
@@ -50,13 +50,11 @@ class EddyMotionEstimator:
         seed : :obj:`int` or :obj:`bool`
             Seed the random number generator (necessary when we want deterministic
             estimation).
-
         Return
         ------
         affines : :obj:`list` of :obj:`numpy.ndarray`
             A list of :math:`4 \times 4` affine matrices encoding the estimated
             parameters of the deformations caused by head-motion and eddy-currents.
-
         """
         align_kwargs = align_kwargs or {}
         reg_target_type = (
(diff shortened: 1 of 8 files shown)


arokem:
@oesteban : I don't see circleci running the tests on this PR. Do we need to change the circleci config somehow to run on PRs?

src/eddymotion/estimator.py
```diff
@@ -105,45 +104,32 @@ def fit(
...
+                        fixed_nii = nb.Nifti1Image(predicted, dwdata.affine)
+                        fixed_ants = ants.from_nibabel(fixed_nii)
+
+                        registration_kwargs = dict(fixed=fixed_ants, moving=moving_ants)
 
-                        registration = Registration(
-                            terminal_output="file",
-                            from_file=pkg_fn(
-                                "eddymotion",
-                                f"config/dwi-to-{reg_target_type}_level{i_iter}.json",
```
arokem, line 123: One question to mull over is how do we pass this config to `ants.registration`? Right now, it's no longer being passed, and it uses the default settings of that function instead.

src/eddymotion/estimator.py
```diff
@@ -105,45 +104,32 @@ def fit(
...
-                            ),
-                            fixed_image=str(fixed.absolute()),
-                            moving_image=str(moving.absolute()),
-                            **align_kwargs,
```
arokem, line 127: Maybe we can use the `align_kwargs` variable to pass the default values in.

src/eddymotion/estimator.py
```diff
@@ -105,45 +104,32 @@ def fit(
...
 
-                        tmpdir = Path(tmpdir)
-                        moving = tmpdir / "moving.nii.gz"
-                        fixed = tmpdir / "fixed.nii.gz"
-                        _to_nifti(data_test[0], dwdata.affine, moving)
-                        _to_nifti(
-                            predicted,
-                            dwdata.affine,
-                            fixed,
-                            clip=reg_target_type == "dwi",
```
arokem, line 116: Looks like the clipping depends on the type of the fixed image (whether it's a b=0 or a diffusion-weighted volume). I guess we need to reincorporate clipping somehow.

src/eddymotion/estimator.py
oesteban: I'm leaning towards just storing one YAML file with the command lines that the estimator can pick.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,23 @@
...
+    val : 
+        Value in dictionary of registration arguments
+
+    Returns
+    -------
+    :obj:`ANTsImage` if val is a string corresponding to a path ending in .nii.gz or .nii
+    """
+    if isinstance(val, str):
+        if Path(val).exists() and (PurePosixPath(val).suffix == '.nii.gz' or PurePosixPath(val).suffix == '.nii'):
+            return ants.from_nibabel(nb.load(val))
```
oesteban, line 21: unnecessary checks. Also, if there's a function to import from nibabel, we don't need to impose ".nii[.gz]" for the extension.

```suggestion
    val = Path(val)
    if not val.exists():
        raise FileNotFoundError(val)
    return ants.from_nibabel(nb.load(val))
```

test/test_utils.py
```diff
@@ -0,0 +1,9 @@
+import ants
+
+
+def test_antsimage_from_path():
```
oesteban, line 4:
```suggestion
def test_antsimage_from_path(datadir, tmp_path):
```

test/test_utils.py
```diff
@@ -0,0 +1,9 @@
+import ants
+
+
+def test_antsimage_from_path():
+    """"""
+    # Generate and save test data to file
+    # Send through antsimage_from_path
+    # Check if ANTsImage, same as original data
+    assert isinstance(result, ants.ANTsImage)
```
oesteban, line 9:
```suggestion
    fixed = tmp_path / "b0.nii.gz"
    dwdata = DWI.from_filename(datadir / "dwi.h5")
    b0nii = nb.Nifti1Image(dwdata.bzero, dwdata.affine, None)
    b0nii.to_filename(fixed)
    assert isinstance(antsimage_from_path(fixed), ants.ANTsImage)
```

src/eddymotion/estimator.py
```diff
@@ -105,45 +107,40 @@ def fit(
...
-                            registration.inputs.initial_moving_transform = str(mat_file)
+                            registration_kwargs['initial_transform'] = str(mat_file)
+
+                        for key in registration_kwargs:
+                            registration_kwargs[key] = antsimage_from_path(registration_kwargs[key])
 
-                        # execute ants command line
-                        result = registration.run(cwd=str(tmpdir)).outputs
+                        #registration_kwargs = ants_utils._int_antsProcessArguments(registration_kwargs)
+                        registration_ants = ants.registration(**registration_kwargs)
```
oesteban, line 133:
```suggestion
                        ants_cli = ants.registration(**registration_kwargs).cmline
                        
                        # Now split ants_cli and replace paths with image objects
                        # this is a bit more convoluted because some paths will come between brackets and commas
                        args = [load_if_image(token) for token in ants_cli.split(" ")]
```

teresamg: Could you show us where the .cmline is coming from? Are you suggesting that we revert back to the nipype Registration() call instead of the current ants call?

arokem:
After some more detailed profiling, we came to the conclusion that this does not provide speedup and potentially reduces flexibility relative to the nipype interface (because the Python API does not expose the full API of the CLI). Unless someone objects, I suggest that we close this PR.

PR #91, opened 2022-11-21
ENH: Add ``FullDKIModel''

oesteban:
The idea behind this model (credit goes to @tmspvn) is to only fit the DKI model once, which could be useful (and, importantly, way faster) for a first pass of head-motion correction.

This new model fits the DKI model at instantiation time, before entering the LOGO loop. Then, the ``predict()`` member just returns the map corresponding to the requested gradient (bvec/bval).

Therefore, it does some "double-dipping" in the sense that the gradient being predicted has already been "seen" in training. The model is likely degraded by the fact that the data are not aligned.

I will keep you updated (cc/ @yasseraleman).

### DIFF
--- a/src/eddymotion/model.py
+++ b/src/eddymotion/model.py
@@ -203,7 +203,7 @@ class TrivialB0Model:
         return self._S0
 
 
-class TrivialDKIModel:
+class TrivialDKIModel(BaseModel):
     """
     A trivial model that returns a *b=0* map always.
 
@@ -219,6 +219,8 @@ class TrivialDKIModel:
     def __init__(self, gtab, data, S0=None, mask=None, b_max=4000, **kwargs):
         """Instantiate the wrapped tensor model."""
         from dipy.reconst.dki import DiffusionKurtosisModel
+
+        super().__init__()
 
         self._S0 = None
         if S0 is not None:
@@ -261,26 +263,10 @@ class TrivialDKIModel:
             # gtab = gtab[:, bval_mask]
 
         self._model = DiffusionKurtosisModel(_rasb2dipy(gtab), **kwargs)
-        self._model = self._model.fit(data[self._mask, ...])
+        super().fit(data, **kwargs)
 
     def fit(self, *args, **kwargs):
         """Do nothing."""
-
-    def predict(self, gradient, step=None, **kwargs):
-        """Propagate model parameters and call predict."""
-
-        predicted = np.squeeze(
-            self._model.predict(
-                _rasb2dipy(gradient),
-                S0=self._S0,
-            )
-        )
-        if predicted.ndim == 3:
-            return predicted
-
-        retval = np.zeros_like(self._mask, dtype="float32")
-        retval[self._mask, ...] = predicted
-        return retval
 
 
 class AverageDWModel:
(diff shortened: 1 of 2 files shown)


src/eddymotion/model.py
```diff
@@ -237,6 +240,95 @@ def predict(self, gradient, **kwargs):
...
+            try:
+                self._model = loop.run_until_complete(asyncio.gather(*fit_tasks))
+            finally:
+                loop.close()
+
+        self._data = _model.predict()
+
+
+    def fit(self, *args, **kwargs):
+        """Do nothing."""
```
arokem, line 325: Do you want a `pass` in here?

src/eddymotion/model.py
```diff
@@ -237,6 +240,95 @@ def predict(self, gradient, **kwargs):
...
+
+    def predict(self, gradient, **kwargs):
+        """Return the *b=0* map."""
+        return self._S0
```
arokem, line 329: This doesn't currently do what the description of the PR says. I think this should return `self._data[..., gradient]` or something like that?

oesteban:
> This should be much (!) faster, right?

Yes, it should be ~N-1 times faster (where N is the number of DWIs).

> The general idea should work well with low-parametric models, maybe even something like SFM but would probably not be so great with a GP, right?

Correct. I think this is a great idea for a fast and efficient approximation to the realignment parameters, possibly in combination with (i) data subsampling; and/or (ii) data smoothing.

Then on a second pass (if time permits, i.e., "production" mode) eddy motion could run the full LOGO modeling with SFM, GP or SHORE, where you may see accuracy gains in avoiding the double-dipping.

oesteban:
(merge #96 first)

src/eddymotion/model.py
oesteban: AFAICT it is not necessary, the docstring definition does the job.

oesteban:
``SingleFitDKIModel`` ?

src/eddymotion/model.py
```diff
@@ -205,6 +208,86 @@ def predict(self, gradient, **kwargs):
         return self._S0
 
 
+class TrivialDKIModel:
```
arokem, line 211: "GlobalDKIModel" ?

"OmnibusDKIModel" ? 

"FullDKIModel" ?

src/eddymotion/model.py
```diff
@@ -205,6 +208,86 @@ def predict(self, gradient, **kwargs):
...
+    """
+    A trivial model that returns a *b=0* map always.
```
arokem, line 213: This docstring needs rewriting.

src/eddymotion/model.py
```diff
@@ -205,6 +208,86 @@ def predict(self, gradient, **kwargs):
...
+
+    Implements the interface of :obj:`dipy.reconst.base.ReconstModel`.
```
arokem, line 215: In particular, maybe we move this text somewhere else?

oesteban:
Okay, this is going in. I'm going to open several follow-up issues gathering some of my own thoughts and some stolen from @tmspvn and @yasseraleman  (I will indicate so :).

But I believe the basic functionality has been fulfilled, hence the merge.

PR #92, opened 2022-11-23
FIX: Housekeeping of the data loader function

oesteban:
This PR revises the data loading:
- Add the possibility to read FSL's bvec/bval files for building the gradient table
- Remove reorientations of the data (still thinking why they were there in the first place)
- Calculating a b=0 map if none is provided

### DIFF
--- a/src/eddymotion/dmri.py
+++ b/src/eddymotion/dmri.py
@@ -235,7 +235,7 @@ def load(
     if b0_file:
         b0img = nb.load(b0_file)
         retval.bzero = np.asanyarray(b0img.dataobj)
-    elif np.any(~gradmsk):
+    elif not np.all(gradmsk):
         retval.bzero = np.median(fulldata[..., ~gradmsk], axis=3)
 
     if brainmask_file:
--- a/src/eddymotion/dmri.py
+++ b/src/eddymotion/dmri.py
@@ -2,6 +2,7 @@
 from collections import namedtuple
 from pathlib import Path
 from tempfile import mkdtemp
+from warnings import warn
 
 import attr
 import h5py
@@ -164,11 +165,15 @@ class DWI:
                         compression_opts=compression_opts,
                     )
 
-    def to_nifti(self, filename):
+    def to_nifti(self, filename, insert_b0=False):
         """Write a NIfTI 1.0 file to disk."""
-        nii = nb.Nifti1Image(self.dataobj, self.affine, None,)
+        data = self.dataobj if not insert_b0 else np.stack((
+            self.bzero[..., np.newaxis], self.dataobj,
+        ))
+        nii = nb.Nifti1Image(data, self.affine, None)
         nii.header.set_xyzt_units("mm")
         nii.to_filename(filename)
+        return
 
     def plot_mosaic(self, index=None, **kwargs):
         """Visualize one direction of the dMRI dataset."""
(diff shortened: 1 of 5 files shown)


src/eddymotion/dmri.py
```diff
@@ -41,10 +41,7 @@ class DWI:
     """
     fieldmap = attr.ib(default=None, repr=_data_repr)
     """A 3D displacements field to unwarp susceptibility distortions."""
-    _filepath = attr.ib(
-        factory=lambda: Path(mkdtemp()) / "em_cache.h5",
-        repr=False,
-    )
+    _filepath = attr.ib(factory=lambda: Path(mkdtemp()) / "em_cache.h5", repr=False,)
```
arokem, line 44: Looks like the black config changed?

src/eddymotion/dmri.py
```diff
@@ -211,35 +199,51 @@ def from_filename(cls, filename):
...
+    bval_file=None,
+    b0_thres=50,
 ):
     """Load DWI data."""
     filename = Path(filename)
     if filename.name.endswith(".h5"):
         return DWI.from_filename(filename)
 
-    if not gradients_file:
+    if gradients_file:
```
arokem, line 216: Want to add some error handling here in case someone foolishly provides both kinds of input?

src/eddymotion/dmri.py
```diff
@@ -211,35 +199,51 @@ def from_filename(cls, filename):
...
+    gradmsk = grad[-1] > b0_thres
     retval.gradients = grad[..., gradmsk]
-    retval.dataobj = img.get_fdata(dtype="float32")[..., gradmsk]
+    retval.dataobj = fulldata[..., gradmsk]
 
     if b0_file:
-        b0img = nb.as_closest_canonical(nb.load(b0_file))
+        b0img = nb.load(b0_file)
         retval.bzero = np.asanyarray(b0img.dataobj)
+    elif np.any(~gradmsk):
```
arokem, line 238:
```suggestion
    elif not np.all(gradmsk):
```

src/eddymotion/dmri.py
```diff
@@ -211,35 +199,51 @@ def from_filename(cls, filename):
...
+        retval.bzero = np.median(fulldata[..., ~gradmsk], axis=3)
```
arokem, line 239: I've never seen anyone use median values as the b0 image before. Is that common?

arokem (commented):
Maybe add some testing of this functionality, while you are at it?

src/eddymotion/dmri.py
oesteban: Yup, I can roll it back if too annoying

src/eddymotion/dmri.py
oesteban: the median is more robust against outliers, but otherwise, it should be reasonably close to the average.

arokem (commented):
Thanks for the added tests! I think this could go in now, but just a couple more comments. Feel free to merge after you've addressed (or if you don't they're necessary to address).

arokem:
Oh - sorry - just noticed that the tests are failing. Good thing they're in place 😄

src/eddymotion/dmri.py
oesteban: Added a warning

src/eddymotion/dmri.py
oesteban: The idea behind this is to have a fallback to generate a b0 quickly. The assumption is that generally, you will provide the `b0_file` argument with a b=0 calculated in a more sophisticated way (e.g., denoising, b1 drift compensation, etc.)

PR #96, opened 2022-11-27
ENH: Improvements to the estimator

oesteban:
Splits the modifications to the estimator from #91.

### DIFF
--- a/src/eddymotion/estimator.py
+++ b/src/eddymotion/estimator.py
@@ -82,32 +82,35 @@ class EddyMotionEstimator:
         for i_iter in range(1, n_iter + 1):
             index_order = np.arange(len(dwdata))
             np.random.shuffle(index_order)
-            with tqdm(total=len(index_order), unit="dwi") as pbar:
-                for i in index_order:
+
+            # Factory creates the appropriate model and pipes arguments
+            dwmodel = ModelFactory.init(
+                gtab=dwdata.gradients,
+                model=model,
+                omp_nthreads=omp_nthreads,
+                **kwargs,
+            )
+
+            with TemporaryDirectory() as tmpdir:
+                print(f"Processing in <{tmpdir}>")
+                with tqdm(total=len(index_order), unit="dwi") as pbar:
                     # run a original-to-synthetic affine registration
-                    with TemporaryDirectory() as tmpdir:
-                        pbar.write(
-                            f"Pass {i_iter}/{n_iter} | Processing b-index <{i}> in <{tmpdir}>"
+                    for i in index_order:
+                        pbar.set_description_str(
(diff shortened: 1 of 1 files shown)


arokem:
This all LGTM. Do you understand the CI test failure?

oesteban:
> Do you understand the CI test failure?

I think I do. We are trying to use asyncio for the parallelization (basically, I stubbornly asked @sebastientourbier about this) and it is failing because the code is likely designed for running inside IPython or have multiple calls to the execution loop.

This made more sense in our first attempt to implement parallelization with processes, but the speed increase was not really so surprising. We then moved into multiple threads and the speedup scaled much better. So, I believe we can just get rid of asyncio for this particular purpose. I'll bring this up later today in our meeting.

oesteban:
Rebased on top of #97, as I assume will resolve the asyncio issue.

src/eddymotion/estimator.py
```diff
@@ -82,32 +82,47 @@ def fit(
         for i_iter in range(1, n_iter + 1):
             index_order = np.arange(len(dwdata))
             np.random.shuffle(index_order)
-            with tqdm(total=len(index_order), unit="dwi") as pbar:
-                for i in index_order:
+
+            dwmodel = None
+            single_model = model.lower() in ("b0", "s0", "avg", "average", "mean")
+            if single_model:
```
arokem, line 88: In this mode, maybe we can predict in one fell swoop and avoid the loop below?

arokem (commented):
Looks great. One thought (see inline) is that in the "single_model" case, we might not even need to loop over gradients, because many of these models have more efficient ways of predicting for all gradients in one fell swoop.

src/eddymotion/estimator.py
oesteban: We still need to run one registration per orientation, so I'm not sure predicting all orientations at once will give us any advantage. The cool thing is that we can now parallelize the loop in the `single_model` case.

This is why I was thinking of thread parallelization of the model fit/predict, and then use multiprocessing parallelization (e.g., joblib, or, better in this case, asyncio) for the loop.

oesteban: BTW, regarding the registration - @tmspvn had a pretty cool idea that would definitely move me away from ANTs (in favor of DIPY where this could be -I believe- implemented):

Very rarely, image registration iterations will be run on the full dataset. We could integrate fit/predict (for costly methods) within the cost function so that we don't have to model all voxels. To have an advantage, we will have the complication of caching modeled voxels as they will likely be picked at some later registration iteration.

WDTY?

oesteban:
I'm going to merge this one and continue the conversations regarding optimizations for new issues that we can schedule better depending on priorities.

PR #97, opened 2022-11-29
FIX: Stop using overly complicated asyncio for managing ThreadPoolExecutors

oesteban:
Rewrite the low-level parallelization of models.


arokem:
Thanks!

PR #103, opened 2022-12-08
FIX: Implement parallelization with joblib & processes

oesteban:
This seems to be working as expected locally. I had forgotten about joblib for a while, and it seems it has evolved very nicely: I was worried about overcommitting omp threads, but it handles it nicely by default :).

### DIFF
--- a/src/eddymotion/model.py
+++ b/src/eddymotion/model.py
@@ -6,14 +6,14 @@ import numpy as np
 from dipy.core.gradients import check_multi_b, gradient_table
 
 
-def _exec_fit(model, data, block=None):
+def _exec_fit(model, data, chunk=None):
     retval = model.fit(data)
-    return retval, block
+    return retval, chunk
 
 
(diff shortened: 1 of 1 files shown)


src/eddymotion/model.py
```diff
@@ -1,21 +1,19 @@
...
-from concurrent.futures import ThreadPoolExecutor, as_completed
-from os import cpu_count
+from joblib import Parallel, delayed
 
 import numpy as np
 from dipy.core.gradients import check_multi_b, gradient_table
 
 
-def _exec_fit(model, data):
+def _exec_fit(model, data, block=None):
```
arokem, line 9: I would prefer "chunk" to "block" because the latter sounds too much like "blocking execution", which is semantically related but not what is intended here.

arokem (commented):
Looks good! Joblib really is great. I have one small comment/suggestion, but other than that I think this is GTG.

PR #110, opened 2022-12-09
ENH: Add SHORE model

oesteban:


oesteban:
This has gotten stale - I'll ping @arokem to learn about wether predict() was finally added to Shore and work from there.

PR #98, opened 2022-11-30
ENH: Outsource leave-one-out splitter so it can be used across data types

teresamg:
(Empty message edited by @oesteban)

This PR covers two targets:

- [x] Extract the "logo" splitter from the `DWI` object. The rationale behind this is allowing its use across different data types (e.g., with PET).
- [x] Incidentally, change the name from "logo" (for leave-one-gradient-out) to "lovo" (for leave-one-volume-out).

PR #112, opened 2022-12-10
ENH: PET uptake model

oesteban:
I still need to test on some data (ha!), but these are the barebones of a PET uptake model.

The idea is a voxelwise fit of a low number of B-Splines (i.e., smooth function) on the data passed into `.fit()`, which can be the full dataset (if initialized with ``FullPETModel``) or all frames but the one being aligned (leave-one-volume-out).

I will try to test on the dataset Martin shared over this week. If that worked, I think we could draft some abstract for OHBM with this and some further validation, WDTY?

cc/ @effigies and @mgxd , who will love to see B-Splines in one more place (it's ironic, but their feedback is always very much appreciated).

I'm attaching @mnoergaard and @arokem as reviewers, but I don't expect feedback until I test this really works on data and remove the "draft" marking of the PR. Feedback before that will be of course welcome nonetheless.


oesteban:
Making progress. @mnoergaard, I'm using the first subject of https://doi.org/10.18112/openneuro.ds004230.v2.3.1 - I don't know if it is a good example.

oesteban:
Okay, I have managed to "simulate" a dataset with this model. The only remaining problems:

- The new PET data structure will need the leave-one-volume-out splitter to be implemented as a function that operates on the data object.
- Fitting the full volume is very slow. Most likely, we don't want to fit one model per voxel.

src/eddymotion/model.py
```diff
@@ -271,6 +270,102 @@ def predict(self, gradient, **kwargs):
...
+
+    def __init__(self, timepoints=None, n_ctrl=None, mask=None, order=3, **kwargs):
+        """
+        Create the B-Spline interpolating matrix.
+
+        Parameters:
+        -----------
+        timepoints : :obj:`list`
+            The timing (in sec) of each PET volume.
+            E.g., ``[20.,   40.,   60.,  120.,  180.,  240.,  360.,  480.,  600.,
```
arokem, line 286: Is it common for PET to be acquired with this kind of non-uniform temporal sampling? (I don't know much about PET, so forgive the potentially naive question)

src/eddymotion/model.py
```diff
@@ -271,6 +270,102 @@ def predict(self, gradient, **kwargs):
...
+        self._x = np.array(timepoints, dtype="float32")
+
+        # Calculate index coordinates in the B-Spline grid
+        self._n_ctrl = n_ctrl or (len(timepoints) // 4) + 1
+
+        # B-Spline knots
+        self._t = np.arange(-3, float(self._n_ctrl) + 4, dtype="float32")
+
+    def fit(self, data, *args, **kwargs):
+        """Do nothing."""
```
arokem, line 310:
```suggestion
        """Fit the model."""
```

src/eddymotion/model.py
```diff
@@ -271,6 +270,102 @@ def predict(self, gradient, **kwargs):
...
+
+        # One single CPU - linear execution (full model)
+        if n_jobs == 1:
+            self._coeff = np.array([
+                cg(ATdotA, AT @ v)[0] for v in data
+            ])
+            return
+
+        # Parallelize process with joblib
+        with Parallel(n_jobs=n_jobs) as executor:
```
arokem, line 342: Do you want to do something for chunking here, instead of doing this with one voxel per task?

src/eddymotion/model.py
```diff
@@ -271,6 +270,102 @@ def predict(self, gradient, **kwargs):
...
+            results = executor(
+                delayed(cg)(ATdotA, AT @ v)
+                for v in data
+            )
+
+        self._coeff = np.array([r[0] for r in results])
+
+    def predict(self, timepoint, **kwargs):
```
arokem, line 350: Random thought here but also relevant to the other models: do we want to raise an error if `predict` is called without first calling `fit`? Potentially store a `.is_fitted_` attribute and check for it, a la sklearn estimators.

arokem (commented):
Looks promising! A couple of random thoughts. In particular, cross-check with #98 for work that @teresamg has been doing to work out memory consumption in the DWI case.

src/eddymotion/data/pet.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # resample and update orientation at index
+        self.dataobj[..., index] = np.asanyarray(
+            xform.apply(dmoving, order=order).dataobj,
+            dtype=self.dataobj.dtype,
+        )
+
+        # update transform
+        if self.em_affines is None:
+            self.em_affines = [None] * len(self)
```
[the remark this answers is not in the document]
oesteban: Yes, the more I work on this, the less value I see in aggregating all these things on a single object if the h5 backing leads to the effect opposite of what is intended.

src/eddymotion/model.py
oesteban: I haven't checked the BIDS specs, but it is my impression that you can define frame starts or frame durations, which are typically nonregular (@mnoergaard is in a much better position to weigh in)

However, your comment has made me realize that the model should be fit at the frame midpoints instead of the starts.

src/eddymotion/model.py
oesteban: I haven't had time to think how to fit several voxels in parallel leveraging numpy. So this is the best I can offer at this point.

src/eddymotion/model.py
oesteban: +1

src/eddymotion/data/pet.py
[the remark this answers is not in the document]
oesteban: As for the data splitter, I will let/help @teresamg toward this

src/eddymotion/model.py
oesteban: Opened #113

oesteban:
Okay, since the data split is not yet outsourced from the DWI object, this model can be tested this way:

```Python
import json
from pathlib import Path
from eddymotion import model
from eddymotion import data

datapath = Path("/data/datasets/ds004230/")

# Load data
metadata = json.loads((datapath / "sub-PS19/ses-baselinebrain/pet/sub-PS19_ses-baselinebrain_rec-DynTOF_pet.json").read_text())
data = pet.load(datapath / "sub-PS19/ses-baselinebrain/pet/sub-PS19_ses-baselinebrain_rec-DynTOF_pet.nii.gz", frame_time=metadata["FrameTimesStart"], frame_duration=metadata["FrameDuration"])

# Initialize and fit model
petmodel = model.PETModel(timepoints=data.frame_time, xlim=data.total_duration)
petmodel.fit(data.dataobj, n_jobs=16)

# Predict fifth frame
predicted = petmodel.predict(data.frame_time[4])
```

mnoergaard:
> Making progress. @mnoergaard, I'm using the first subject of https://doi.org/10.18112/openneuro.ds004230.v2.3.1 - I don't know if it is a good example.

Hi @oesteban! This is great - thanks for pushing this forward! ds004230 should be an ideal candidate with "only" a size of about 160MB per PET file. The largest PET files I have come across are around 1GB (high resolution scanners). For ds004230, you can also find the current motion correction results for comparison in /derivatives/petsurfer/logs/. The generated confounds file is located in /derivatives/petsurfer/subXX/sesXX/pet/subXX-sesXX_desc-confounds_timeseries.tsv. Furthermore, motion free PET data is also generated/synthesized and can be found here /derivatives/petsurfer/subXX/sesXX/pet/nopvc/yhat.nii.gz. 

This code has been used to perform motion correction for these data (https://github.com/mnoergaard/hmcpet/blob/master/init_pet_hmc_wf.py), that includes both a selection of frames (only frames after 2 minutes, because there is no signal before this) and also a smoothing step (10 mm) to increase SNR. These could ideally be included in the eddymotion workflow? 

Let me know if we should touch base over zoom soon.

src/eddymotion/model/base.py
```diff
@@ -293,6 +292,107 @@ def predict(self, gradient, **kwargs):
...
+        self._shape = data.shape[:3]
+
+        # Convert data into V (voxels) x T (timepoints)
+        data = (
+            data.reshape((-1, data.shape[-1]))
+            if self._mask is None else data[self._mask]
+        )
+
+        # A.shape = (T, K - 4); T= n. timepoints, K= n. knots (with padding)
+        A = BSpline.design_matrix(x, self._t, k=self._order)
```
mnoergaard, line 356: The design_matrix in BSpline was not introduced in scipy until v 1.8.0, so should require >= than this in the setup.cfg (https://scipy.github.io/devdocs/reference/generated/scipy.interpolate.BSpline.design_matrix.html).

src/eddymotion/model.py
mnoergaard: Yes - I would say this is quite common (for dynamics PET with a bolus injection), but there are also cases where uniform sampling is used (infusion studies).

mnoergaard (commented):
Thanks for pushing these efforts forward @oesteban - I think this looks pretty good so far!

oesteban:
I think I will merge this as a base, and then we can address the frame selection by allowing weights (and fitting with WLS).

Issue #107, opened 2022-12-09
Show a big-fat warning when a brain mask is not provided


oesteban:
With the development of #91, it has become clear that we don't want to be fitting models in background pixels, nor have the background be considered by the cost function of ANTs.

Therefore, eddymotion should be very reluctant to go ahead without a brain mask.

Incidentally, I believe can easily include SynthStrip into dMRIPrep and then dilate that mask so that distortions do not cause cropping of the brain.

WDYT?

arokem:
I ❤️ synthstrip and I think that's a great idea. We should also give the option to provide masks from other pipelines.

oesteban:
Correct, we don't want to mess with how the user calculates masks - eddymotion should just trust the user is doing the right thing on this aspect.

PR #114, opened 2022-12-13
Fix masking add composability

dPys:
@oesteban -- It is truly incredible how far this repo has come. I also still believe it remains right on the cusp of making eddy correction of dmri more accessible to researchers looking for other alternatives to FSL. 

There's very little to complain about with how you've organized things already. With that said, I do have some proposed enhancements, any or none of which you're welcome to adopt. I realize that this first PR comes after a long lull in my contribution, so it mainly just includes some general housekeeping, most of which comes from debugging locally with test data that includes a b0 mask. Here is a brief summary of the included changes:

* Relegates registration logic previously belonging to the `EddyMotionEstimator` class to a more modular Aligner dataclass. This may ultimately also make it easier to surgically swap out nipype-based native ANTs with ANTspy, if that strategy is still on the agenda per #89 and #90 .
* Refactors the order and methodology of parsing `S0` and `mask`  from kwargs to address cases where either or both are manually provided by the user. In the case of a B0 mask, for example, I found that np.ma.masked_array behaves more predictably than simple slicing. These use-cases probably also need to further unit tests before this PR gets merged. Per #107 , it might also be worth proceeding to raise a warning if a mask is not provided.
* Adds some additional runtime error handling and simplified results concatenation logic around the joblib calls in the fit/predict logic of `BaseModel`.
* Collectively, these improvements should help to shave off some memory consumption and lead to a small speedup overall.
* My linter auto-reformatted some things per black as well as flake8, so feel free to reject those changes if you have a different formatting aesthetic in mind.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -2,8 +2,8 @@ Anisha Keshavan <akeshavan@gmail.com> <akeshavan@ucla.edu>
 Anisha Keshavan <akeshavan@gmail.com> <keshavan@berkeley.edu>
 Ariel Rokem <arokem@berkeley.edu>
 Ariel Rokem <arokem@berkeley.edu> <arokem@gmail.com>
-Derek Pisner <dpisner@utexas.edu>
-Derek Pisner <dpisner@utexas.edu> dPys <dpisner@utexas.edu>
+Derek Pisner <dpysalexander@gmail.com>
+Derek Pisner <dpysalexander@gmail.com> dPys <dpysalexander@gmail.com>
 Garikoitz Lerma-Usabiaga <garikoitz@gmail.com> garikoitz <garikoitz@gmail.com>
 James D. Kent <james-kent@uiowa.edu>
 James D. Kent <james-kent@uiowa.edu> <jamesdkent21@gmail.com>
--- a/.mailmap
+++ b/.mailmap
@@ -2,8 +2,8 @@ Anisha Keshavan <akeshavan@gmail.com> <akeshavan@ucla.edu>
 Anisha Keshavan <akeshavan@gmail.com> <keshavan@berkeley.edu>
 Ariel Rokem <arokem@berkeley.edu>
 Ariel Rokem <arokem@berkeley.edu> <arokem@gmail.com>
-Derek Pisner <dpysalexander@gmail.com>
-Derek Pisner <dpysalexander@gmail.com> dPys <dpysalexander@gmail.com>
+Derek Pisner <dpisner@utexas.edu>
+Derek Pisner <dpisner@utexas.edu> dPys <dpisner@utexas.edu>
 Garikoitz Lerma-Usabiaga <garikoitz@gmail.com> garikoitz <garikoitz@gmail.com>
 James D. Kent <james-kent@uiowa.edu>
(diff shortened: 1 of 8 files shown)


.mailmap
```diff
@@ -2,8 +2,8 @@ Anisha Keshavan <akeshavan@gmail.com> <akeshavan@ucla.edu>
 Anisha Keshavan <akeshavan@gmail.com> <keshavan@berkeley.edu>
 Ariel Rokem <arokem@berkeley.edu>
 Ariel Rokem <arokem@berkeley.edu> <arokem@gmail.com>
-Derek Pisner <dpisner@utexas.edu>
-Derek Pisner <dpisner@utexas.edu> dPys <dpisner@utexas.edu>
+Derek Pisner <dpysalexander@gmail.com>
+Derek Pisner <dpysalexander@gmail.com> dPys <dpysalexander@gmail.com>
```
oesteban, line 6: This will remove your name from old commits done with the old email.

That said, please send a separate PR with this edit.
```suggestion
Derek Pisner <dpysalexander@gmail.com> dPys <dpysalexander@gmail.com>
Derek Pisner <dpysalexander@gmail.com> <dpisner@utexas.edu> 
```

setup.cfg
```diff
@@ -1,7 +1,7 @@
 [metadata]
 name = eddymotion
 author = Derek Pisner
-author_email = dpisner@utexas.edu
+author_email = dpysalexander@gmail.com
```
oesteban, line 4: Like the other, please pull out on the other PR. I think we should also follow the general nipreps guidelines and write the nipreps account.

```suggestion
author_email = nipreps@gmail.com
```

setup.cfg
```diff
@@ -15,7 +15,7 @@ license = Apache License, Version 2.0
 long_description = file:README.rst
 long_description_content_type = text/x-rst; charset=UTF-8
 maintainers = The NiPreps developers
-maintainer_email = nipreps@gmail.com, dpisner@utexas.edu
+maintainer_email = nipreps@gmail.com, dpysalexander@gmail.com
```
oesteban, line 18: Ditto.

oesteban (commented):
I need to revise the PR, but for the time being, please separate the packaging/metadata changes.

.mailmap
dPys: See https://github.com/nipreps/eddymotion/pull/119

setup.cfg
dPys: Ditto, see https://github.com/nipreps/eddymotion/pull/119

setup.cfg
dPys: see https://github.com/nipreps/eddymotion/pull/119

dPys:
No rush, re: any revisions.

Not sure how my contact info updates slipped into my final commit-- sorry about that!  RE: packaging/metadata changes, I do actually have some more related to `pyproject.toml` specifically, but will scope out those commits to a third PR once this one and https://github.com/nipreps/eddymotion/pull/119 are over the finish-line.

oesteban:
We have worked several aspects of this PR elsewhere, so I think it's safe to close. Thanks to @dPys for the effort - we've learned a lot from this.

Issue #108, opened 2022-12-09
Spatial smoothing & disallowing ANTs' subsamplings/smoothings


oesteban:
The first registration level (likely using the new `FullDKIModel` or similar) can substantially be sped up if we spatially smoothed & downsampled the input before both model fitting and image registration.

This will make ANTs' settings easier, as the registration pyramid is set up globally, not just for registration.

This is one of the ideas I partially stole from @tmspvn.

arokem:
That's a 💯 idea!

PR #116, opened 2022-12-15
ENH: Data manipulations to facilitate modeling and registration: smoothing & subsampling

oesteban:
Resolves: #108

### DIFF
--- a/src/eddymotion/data/filtering.py
+++ b/src/eddymotion/data/filtering.py
@@ -23,6 +23,93 @@
 """Filtering data."""
 
 
+def gaussian_filter(data, vox_width):
+    """
+    Apply a Gaussian smoothing filter of a given width (in voxels)
+
+    Parameters
+    ----------
+    data : :obj:`numpy.ndarray`
+        The input image's data array
+    vox_width : :obj:`numbers.Number` or :obj:`tuple` or :obj:`list`
+        The smoothing kernel width in voxels
+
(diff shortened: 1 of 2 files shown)


esavary:
@oesteban This PR could be useful for Gaussian processes. Is it ready for us to take over and work on it?

oesteban:
This would need to add tests. I'm preparing some data to do that.

src/eddymotion/data/filtering.py
```diff
@@ -0,0 +1,236 @@
...
+
+    # Create new Nifti image with updated information
+    newnii = Nifti1Image(resampled, newaffine, imnii.header)
+    newnii.set_sform(newaffine, code=1)
+    newnii.set_qform(newaffine, code=1)
+
+    return newnii
+
+
+def advanced_clip(
```
mnoergaard, line 176: For PET, we usually do this clipping on a frame-wise basis, because the signal changes so much from frame to frame.

mnoergaard: Arh, I see - this also expects a 3D input.

oesteban: This is done by frame, yes -- this one should not be a problem (although, I see the point in also supporting global clipping).

oesteban:
This PR carries too much weight with it. I'm going to split it in two (potentially three).

PR #120, opened 2022-12-16
MAINT: Pin Python >= 3.8

oesteban:
Although Python 3.7 end-of-life is set for in 6 months (Jun 27, 2023), scikit-learn does not support it as of 1.8.0 (which will become the minimal version with #161).

I belive discontinuing Python 3.7 at this point is better than implementing the bspline matrix ourselves.


oesteban:
I meant scipy >= 1.8.0 above.

PR #122, opened 2023-01-05
Changed em_affines from Affine object to matrix with preallocation

teresamg:


arokem:
This looks correct to me. For future reference, this implementation is far more memory efficient than storing the affines in a list, which is why this change is needed. The list implementation was gobbling up memory at a terrifying speed, at least on some machines.

Anothe note is that IIUC, formatting changes are to enforce black formatting.

arokem:
We also need to change the test accordingly, to avoid this error: https://app.circleci.com/pipelines/github/nipreps/eddymotion/285/workflows/80b68fb0-dda7-4917-960e-9d1667b604b6/jobs/272?invite=true#step-110-571

test/test_integration.py
```diff
@@ -95,4 +95,5 @@ def test_proximity_estimator_trivial_model(datadir, tmp_path):
     coords = xfms.reference.ndcoords.T
     for i, est in enumerate(em_affines):
         xfm = nt.linear.Affine(xfms.matrix[i], reference=b0nii)
+        est = nt.linear.Affine(est.matrix, reference=b0nii)
```
arokem, line 98: Looks like this is still failing (https://app.circleci.com/pipelines/github/nipreps/eddymotion/286/workflows/7d9d8e2b-f6da-4968-8c21-1b0fbb3775dc/jobs/273). I believe it should be: 
```suggestion
        est = nt.linear.Affine(est, reference=b0nii)
```

src/eddymotion/data/dmri.py
```diff
@@ -167,9 +167,9 @@ def set_transform(self, index, affine, order=3):
 
         # update transform
         if self.em_affines is None:
-            self.em_affines = [None] * len(self)
+            self.em_affines = np.zeros((self.dataobj.shape[-1], 4, 4))
 
-        self.em_affines[index] = xform
+        self.em_affines[index] = xform._matrix
```
oesteban, line 172:
```suggestion
        self.em_affines[index] = xform.matrix
```

src/eddymotion/estimator.py
```diff
@@ -182,9 +187,23 @@ def fit(
                         if bmask_img:
                             registration.inputs.fixed_image_masks = ["NULL", bmask_img]
 
-                        if dwdata.em_affines and dwdata.em_affines[i] is not None:
+                        if dwdata.em_affines is not None and np.any(
+                            dwdata.em_affines[i]
+                        ):
```
oesteban, line 192:
```suggestion
                        if np.any(dwdata.em_affines[i, ...]):
```

test/test_integration.py
oesteban: Yes, `est` is a 4x4 ndarray, not an Affine object.

oesteban (commented):
looking good

arokem:
With tests now passing, I think it's ready to merge!

oesteban:
Good job @teresamg !

PR #128, opened 2024-03-12
MAINT: Migrate to PEP517/518 packaging - WIP Do not merge

esavary:


pyproject.toml
```diff
@@ -1,15 +1,180 @@
...
+    "nbsphinx",
+    "packaging",
+    "pydot >= 1.2.3",
+    "pydotplus",
+    "sphinx >= 2.1.2",
+    "sphinx-argparse",
+    "sphinx_rtd_theme",
+    "sphinxcontrib-apidoc ~= 0.3.0",
+    "sphinxcontrib-napoleon",
+    "sphinxcontrib-versioning"
```
jhlegarreta, line 52: `furo` needs to be added here: https://github.com/nipreps/eddymotion/blob/4b632ea345faa69088c03447dc9e707d93c62c32/docs/conf.py#L115

effigies:
Apologies for spinning this out. It kind of became one thing after another, as the installation shows up in several places that weren't being tested properly.

This now:

1) Moves to hatch/hatch-vcs.
2) Updates CI in various places and ensures docs build on PRs.
3) Fixes doc build and lint issues.
4) Tests up to Python 3.12, which was probably an overreach, but I was in there already.

effigies (approved):
This is ready to go AFAIC.

.maint/update_requirements.py
oesteban, line 1: This is not used anywhere, highly likely transferred from sdcflows as it was the example I provided. Can we remove it?

requirements.txt
oesteban, line 1: Like `update_requirements.py`, this is copied from sdcflows. While it has a role there, I don't think we need it here for now. Can we remove it?

oesteban (commented):
Just added two minimal comments, which came from the fact that I gave a probably not-ideal example reference PR with sdcflows.

.maint/update_requirements.py
esavary: Yes, indeed. I noticed that an update_requirements.py script was added to generate requirements.txt file in nipreps/sdcflows#410 , so I've done something similar. I've removed the files in my last commit, can I proceed with merging?

docs/requirements.txt
```diff
@@ -1,6 +1,7 @@
 attrs >= 20.1.0
-furo ~= 2021.10.09
+furo >= 2024.01.29
```
oesteban, line 2: I'm not sure this change will work -- let's keep an eye on this. I remember pinning furo because of some weird dependency failing in more recent builds. Perhaps they figured out.

oesteban (approved):
Please feel free to merge when you're ready @esavary

docs/requirements.txt
effigies: I built it locally.

PR #129, opened 2024-03-15
ENH: Add newer Python versions to package testing

jhlegarreta:
Add newer Python versions to package testing: add Python 3.11 and Python 3.12.


jhlegarreta:
To be considered after #128 has been merged.

PR #133, opened 2024-03-15
ENH: Define instance attributes in `__init__` method

jhlegarreta:
Define instance attributes in `__init__` method.

Fixes:
```
Instance attribute _datashape defined outside __init__
```

and similar warnings.

### DIFF
--- a/src/eddymotion/model/base.py
+++ b/src/eddymotion/model/base.py
@@ -131,6 +131,9 @@ class BaseModel:
         module_name, class_name = model_str.rsplit(".", 1)
         self._model = getattr(import_module(module_name), class_name)(_rasb2dipy(gtab), **kwargs)
 
+        self._datashape = None
+        self._models = None
+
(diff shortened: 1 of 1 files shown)


oesteban:
I'm surprised - is this a problem despite them being listed in ``__slots__``?

Happy to merge if this is not just the face of a deeper problem

jhlegarreta:
Looks like it is still recommended to initialize the values in the `__init__` method:
https://wiki.python.org/moin/UsingSlots

If the above ref is not enough, the below does not say they are a substitute for the `__init__` initialization:
https://docs.python.org/3/reference/datamodel.html#object.__slots__

And `pylint` has a warning for this as well:
https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/attribute-defined-outside-init.html

Not sure whether there are other aspects that make the behavior to be different in some cases.

PR #134, opened 2024-03-15
ENH: Remove unused parameter from method prototype

jhlegarreta:
Remove unused parameter from method prototype: PET images do not have a `b0` image. Use the `*_` to denote unused arguments for the `PET` class and use a `kwargs` dictionary for the `DWI` case.

### DIFF
--- a/src/eddymotion/data/dmri.py
+++ b/src/eddymotion/data/dmri.py
@@ -195,8 +195,9 @@ class DWI:
                         compression_opts=compression_opts,
                     )
 
-    def to_nifti(self, filename, insert_b0=False):
+    def to_nifti(self, filename, **kwargs):
         """Write a NIfTI 1.0 file to disk."""
+        insert_b0 = kwargs.get("insert_b0", False)
         data = (
             self.dataobj
             if not insert_b0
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Not sure whether this was put intentionally to make the interface the same as for `DWI` but in that case a base class would be more appropriate, or to use kwargs.

oesteban:
This is intentional, perhaps a bad way of defining an API, but all methods need to have that argument :)

How would you go about it?

jhlegarreta:
My first thought is that it should not be there if it is not common to all classes :-S

Thoughts:
- Could be a passed as kwargs and do `insert_b0 = kwargs['insert_b0']` for the DWI case.
- Other solutions:
  - Mark as  `# pylint: disable=unused-argument` or `# noqa: F841` (less preferable than the below)
  - Do `_ = (insert_b0)` as the first statement in the method in the PET case
  - Do `del insert_b0` as the first statement in the method in the PET case
  - Declare the method as `def to_nifti(self, filename, *_):` in the PET case
  - Declare the method as `def to_nifti(self, filename, *insert_b0):` in the PET case

oesteban:
> * Do `_ = (insert_b0)` as the first statement in the method in the PET case

This argument is strong - why PET has to have `insert_b0` in first place. Your solution of using `kwargs` seems the right choice.

Thanks for point out this issue.

jhlegarreta:
Done in 102a820.

oesteban (approved):

PR #136, opened 2024-03-18
BUG: Fix Sphinx documentation configuration file

jhlegarreta:
- BUG: Specify the language in Sphinx documentation config file
- BUG: Fix DIPY doc URL in Sphinx documentation config file

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -91,7 +91,7 @@ numfig = True
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -235,7 +235,7 @@
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     "bids": ("https://bids-standard.github.io/pybids/", None),
```
jhlegarreta, line 237: All versions pointed here should probably be synchronized with the versions specified when listing the dependencies, but that is a separate issue/PR.

jhlegarreta:
BTW, the `_version.py` file is not being generated, contrary to what is said in
https://github.com/nipreps/eddymotion/blob/4b632ea345faa69088c03447dc9e707d93c62c32/.github/workflows/docs-build-update.yml#L50

and
https://github.com/nipreps/eddymotion/blob/4b632ea345faa69088c03447dc9e707d93c62c32/.github/workflows/docs-build-pr.yml#L47

I had to call `python setup.py` to set up the package and create the file. That is why GitHub actions are reporting
```
ModuleNotFoundError: No module named 'eddymotion._version'
```

e.g. https://github.com/nipreps/eddymotion/actions/runs/8326718057/job/22782937133#step:6:36

Not sure if skipping that step was intentional. The call will change anyways after #128.

PR #139, opened 2024-03-21
CI: replace flake8 by ruff check

esavary:
- Replace flake8 CI job with a ruff CI job.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,9 +148,7 @@ select = [
   "B",
 ]
 ignore = [
-  "N802",
-  "N806",
-  "E203",
+  "E203"
 ]
 
 [tool.ruff.lint.flake8-quotes]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,6 +158,7 @@ inline-quotes = "double"
 "*/__init__.py" = ["F401"]
 "docs/conf.py" = ["E265"]
 "/^\\s*\\.\\. _.*?: http/" = ["E501"]
+"src/eddymotion/estimator.py" = ["C901"]
 
 [tool.ruff.format]
 quote-style = "double"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -140,7 +140,7 @@ skip_gitignore = true
 line-length = 99
 
 [tool.ruff.lint]
-extend-select = [
+select = [
   "F",
   "E",
   "C",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,7 +146,6 @@ extend-select = [
   "C",
   "W",
   "B",
-  "B950",
 ]
 ignore = [
   "N802",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,46 +143,28 @@ line-length = 99
 extend-select = [
   "F",
   "E",
+  "C",
   "W",
-  "I",
-  "UP",
-  "YTT",
-  "S",
-  "BLE",
   "B",
-  "A",
-  # "CPY",
-  "C4",
-  "DTZ",
-  "T10",
-  # "EM",
-  "EXE",
-  "FA",
-  "ISC",
-  "ICN",
-  "PT",
-  "Q",
+  "B950",
 ]
 ignore = [
-  "S311",  # We are not using random for cryptographic purposes
-  "ISC001",
-  "S603",
+  "N802",
+  "N806",
+  "W503",
+  "E203",
 ]
 
 [tool.ruff.lint.flake8-quotes]
-inline-quotes = "single"
+inline-quotes = "double"
 
 [tool.ruff.lint.extend-per-file-ignores]
-"*/test_*.py" = ["S101"]
-"docs/source/conf.py" = ["A001"]
-"mriqc/bin/nib_hash.py" = ["S324"]
-"mriqc/config.py" = ["S105"]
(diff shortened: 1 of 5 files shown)


esavary:
In pyproject.toml, I'm wondering how to choose the error codes for the section [tool.ruff.lint]. Should we use the same as fmriprep or mriqc, or create specific rules for eddymotion? What are your thoughts on this @effigies @oesteban ?

oesteban:
Except for `src/eddymotion/estimator.py:43:9: C901 `fit` is too complex (17 > 10)`, I don't think we want to have exceptions there. Copying from fmriprep/mriqc does not seem useful at all -- you want to be very explicit about what you add as an exception.

And maybe we should look into that C901 and try to pacify ruff by improving the code, as opposed to adding one exception.

esavary:
> Except for `src/eddymotion/estimator.py:43:9: C901 `fit` is too complex (17 > 10)`, I don't think we want to have exceptions there. Copying from fmriprep/mriqc does not seem useful at all -- you want to be very explicit about what you add as an exception.
> 
> And maybe we should look into that C901 and try to pacify ruff by improving the code, as opposed to adding one exception.

More explicitly, are you suggesting adding B028 and C408 to the list of errors to ignore and modifying estimator.py to address C901?

oesteban:
> > Except for `src/eddymotion/estimator.py:43:9: C901 `fit` is too complex (17 > 10)`, I don't think we want to have exceptions there. Copying from fmriprep/mriqc does not seem useful at all -- you want to be very explicit about what you add as an exception.
> > And maybe we should look into that C901 and try to pacify ruff by improving the code, as opposed to adding one exception.
> 
> More explicitly, are you suggesting adding B028 and C408 to the list of errors to ignore and modifying estimator.py to address C901?

I think we should not have exceptions. We may add C901 temporarily so that this PR passes but open an issue to remind us of looking into it at a later time.

esavary:
> > > Except for `src/eddymotion/estimator.py:43:9: C901 `fit` is too complex (17 > 10)`, I don't think we want to have exceptions there. Copying from fmriprep/mriqc does not seem useful at all -- you want to be very explicit about what you add as an exception.
> > > And maybe we should look into that C901 and try to pacify ruff by improving the code, as opposed to adding one exception.
> > 
> > 
> > More explicitly, are you suggesting adding B028 and C408 to the list of errors to ignore and modifying estimator.py to address C901?
> 
> I think we should not have exceptions. We may add C901 temporarily so that this PR passes but open an issue to remind us of looking into it at a later time.

Can you provide more details? If I leave out B028 and C408 from the exceptions, the PR won't pass. Right now, I've used error codes from the old Flake8 configuration for the extend and ignore lists in [tool.ruff.lint] in the pyproject.toml, but I'm not sure if this is the most effective approach.

pyproject.toml
```diff
@@ -136,18 +136,34 @@ exclude = '''
...
+  "F",
+  "E",
+  "C",
+  "W",
+  "B",
+]
+ignore = [
+  "N802",
+  "N806",
+  "E203",
```
oesteban, line 153: I think we can live with [E203](https://docs.astral.sh/ruff/rules/whitespace-before-punctuation/) here.
```suggestion
  "E203"
```

Let's break this PR with the other two and see where we are using bad function/variable names.

pyproject.toml
```diff
@@ -136,18 +136,34 @@ exclude = '''
...
 ]
 
+[tool.ruff.lint.flake8-quotes]
+inline-quotes = "double"
+
+[tool.ruff.lint.extend-per-file-ignores]
+"*/__init__.py" = ["F401"]
+"docs/conf.py" = ["E265"]
+"/^\\s*\\.\\. _.*?: http/" = ["E501"]
+
```
oesteban, line 163: I feel it is preferable to avoid constant excludes and use per-file policies. Here, we add C901 only for a particular file where it was found:

```suggestion
"src/eddymotion/estimator.py" = ["C901"]

```

oesteban:
Okay, we have this at the moment:

```
src/eddymotion/data/dmri.py:254:13: B028 No explicit `stacklevel` keyword argument found
src/eddymotion/model/base.py:93:18: C408 Unnecessary `tuple` call (rewrite as a literal)
src/eddymotion/viz.py:346:24: C408 Unnecessary `dict` call (rewrite as a literal)
src/eddymotion/viz.py:398:22: C408 Unnecessary `dict` call (rewrite as a literal)
```

These are trivial changes we can (and should) address.

esavary:
> Okay, we have this at the moment:
> 
> ```
> src/eddymotion/data/dmri.py:254:13: B028 No explicit `stacklevel` keyword argument found
> src/eddymotion/model/base.py:93:18: C408 Unnecessary `tuple` call (rewrite as a literal)
> src/eddymotion/viz.py:346:24: C408 Unnecessary `dict` call (rewrite as a literal)
> src/eddymotion/viz.py:398:22: C408 Unnecessary `dict` call (rewrite as a literal)
> ```
> 
> These are trivial changes we can (and should) address.

Should this be addressed in this PR, or would it be more appropriate for a separate PR? (Considering the PR subject is transitioning from flake8 to ruff)

oesteban:
It's only four little changes -- I would stuff it in this one and reduce the noise overall

oesteban (approved):
LGTM

PR #141, opened 2024-03-24
ENH: Use NumPy's `Generator` class as a replacement for `RandomState`

jhlegarreta:
Use NumPy's `Generator` class as a replacement for `RandomState` for random variarte generation methods.

`RandomState` was deprecated in NumPy 1.17.0.

Take advantage of the commit to document better the use of the `seed` parameter.

Documentation:
https://numpy.org/doc/stable/reference/random/new-or-different.html#what-s-new-or-different
https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html

### DIFF
--- a/src/eddymotion/estimator.py
+++ b/src/eddymotion/estimator.py
@@ -75,8 +75,9 @@ class EddyMotionEstimator:
             Number of parallel jobs.
         seed : :obj:`int` or :obj:`bool`
             Seed the random number generator (necessary when we want deterministic
-            estimation).
-
+            estimation). If an integer, the value is used to initialize the
+            generator; if ``True``, the arbitrary value of ``20210324`` is used
+            to initialize it.
         Return
         ------
         affines : :obj:`list` of :obj:`numpy.ndarray`
@@ -86,8 +87,11 @@ class EddyMotionEstimator:
         """
         align_kwargs = align_kwargs or {}
 
+        _seed = None
         if seed or seed == 0:
-            np.random.seed(20210324 if seed is True else seed)
+            _seed = 20210324 if seed is True else seed
+
+        rng = np.random.default_rng(_seed)
 
         if "num_threads" not in align_kwargs and omp_nthreads is not None:
             align_kwargs["num_threads"] = omp_nthreads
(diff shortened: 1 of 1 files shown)


src/eddymotion/estimator.py
```diff
@@ -86,8 +86,11 @@ def fit(
         """
         align_kwargs = align_kwargs or {}
 
+        _seed = None
         if seed or seed == 0:
-            np.random.seed(20210324 if seed is True else seed)
+            _seed = 20210324 if seed is True else seed
```
arokem, line 91: While we're here, can you add the default behavior to the docstring? IIUC, the default is to use `20210324` as the seed?

jhlegarreta: > While we're here, can you add the default behavior to the docstring? 

Done in d77d289.

src/eddymotion/estimator.py
```diff
@@ -75,7 +75,8 @@ def fit(
             Number of parallel jobs.
         seed : :obj:`int` or :obj:`bool`
```
oesteban, line 76:
```suggestion
        seed : :obj:`int` or :obj:`bool` (default: ``20210324``)
```

src/eddymotion/estimator.py
```diff
@@ -75,7 +75,8 @@ def fit(
...
             Seed the random number generator (necessary when we want deterministic
-            estimation).
+            estimation). If an integer, the value is used to initialize the
+            generator; if True, an internal value is used to initialize it.
```
oesteban, line 79:
```suggestion
            generator; if ``True`` or ``0``, the arbitrary value of ``20210324`` is used to initialize it.
```

oesteban (approved):
Just a nitpick following @arokem's suggestion

src/eddymotion/estimator.py
jhlegarreta: Done in b0e53e4. Generally, I do not think adding hard-coded values to the docstring is a good practice: the risk of changing the value for some reason and forgetting to update the docstring is usually high. Best would be to declare the value in some utils file, use the value throughout the code, and cross-reference it in the docstrings.

For the time being, I have changed the docstring as suggested though.

oesteban: That's a good point. This value should be a static variable of the module with its own docstring.

PR #144, opened 2024-03-27
ENH: Refactor code linter and formatter dependencies

jhlegarreta:
Refactor code linter and formatter dependencies:
- Remove `black`, `flake8` and `isort` dependencies from `pyproject.toml`: the role has been taken over by `ruff` in commit 81b3690.
- Transfer the `black` and `isort` rules left behind to `ruff` in `pyproject.toml`
- Remove `black`, `flake8` and `isort` from the pre-commit config file and use `ruff` instead.

### DIFF
--- a/.github/workflows/contrib.yml
+++ b/.github/workflows/contrib.yml
@@ -20,4 +20,4 @@ jobs:
         with:
           python-version: 3
       - name: Lint EddyMotion
-        run: pipx run ruff check
+        run: pipx run ruff check --diff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,15 +10,9 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: debug-statements
-  - repo: https://github.com/timothycrosley/isort
-    rev: 5.10.1
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.3.4
     hooks:
-      - id: isort
-  - repo: https://github.com/psf/black
-    rev: 22.3.0
-    hooks:
-      - id: black
-  - repo: https://github.com/pycqa/flake8
-    rev: 4.0.1
-    hooks:
-      - id: flake8
+      - id: ruff
+        args: [ --fix ]
+      - id: ruff-format
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -6,8 +6,10 @@ full list see the documentation:
 http://www.sphinx-doc.org/en/master/config
 
 """
+
 from packaging.version import Version
-from eddymotion import __version__, __copyright__, __packagename__
+
+from eddymotion import __copyright__, __packagename__, __version__
 
 # -- Path setup --------------------------------------------------------------
 # If extensions (or modules to document with autodoc) are in another directory,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,10 +57,7 @@ doc = [
 ]
 
 dev = [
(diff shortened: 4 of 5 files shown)


jhlegarreta:
A few comments:
- There seems to be a contradiction between the use of `skip-string-normalization = true` in
  https://github.com/nipreps/eddymotion/blob/main/pyproject.toml#L118
  and `inline-quotes = "double"` in
  https://github.com/nipreps/eddymotion/blob/main/pyproject.toml#L155

  according to:
  https://github.com/astral-sh/ruff/pull/8822

  `skip-string-normalization = true` is `quote-style=preserve` ruff, but `inline-quotes = "double"` enforces particular style.

- Looks like this adding `ruff`'s import sorting:
  https://github.com/nipreps/eddymotion/pull/144/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R135

  slightly changes `isort`'s vs `ruff`'s import sorting:
  https://github.com/nipreps/eddymotion/actions/runs/8458933073/job/23174220049?pr=144#step:4:14

  Their rules are known to be different:
  https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort

  Do we want `I001` to be excluded?

- On a related note, before the cleanup proposed by this PR I was getting an error related to `isort` locally:
  ```
  [INFO] Installing environment for https://github.com/timothycrosley/isort.
  [INFO] Once installed this environment will be reused.
  [INFO] This may take a few minutes...
  (...)
    × Preparing metadata (pyproject.toml) did not run successfully.
    │ exit code: 1
     ╰─> [14 lines of output]
          Traceback (most recent call last):
  (...)
              raise RuntimeError("The Poetry configuration is invalid:\n" + message)
          RuntimeError: The Poetry configuration is invalid:
            - data.extras.pipfile_deprecated_finder[2] must match pattern ^[a-zA-Z-_.0-9]+$

        [end of output]

   note: This error originates from a subprocess, and is likely not a problem with pip.
  error: metadata-generation-failed
  ```

  which turns out to be related to the specific `isort` version that is being used. Removing `isort` fixes it, as `ruff` has taken over.

jhlegarreta:
Also, we may want to add the `--fix` option and `ruff format` in here:
https://github.com/nipreps/eddymotion/blob/d23dcf02a2cd5bc8965bfd294cebbe6216fe8373/.github/workflows/contrib.yml#L23

so that it tells us how it would fix the code, instead of trying to discover it by ourselves.

oesteban:
> Also, we may want to add the `--fix` option and `ruff format` in here:
> 
> https://github.com/nipreps/eddymotion/blob/d23dcf02a2cd5bc8965bfd294cebbe6216fe8373/.github/workflows/contrib.yml#L23
> 
> so that it tells us how it would fix the code, instead of trying to discover it by ourselves.

I don't think we would see how to fix the code unless we do a git diff right after, but I'm happy to do that.

oesteban:
Oh, I see, this is not the github action, it's pre-commit.

I personally would prefer to avoid pre-commit. But it is not a hill I intend to die on.

jhlegarreta:
Oh, yes, looks like meant more a diff-like thing than a fix for GHA, and the pre-commit config is probably only checking the modified files, and thus we are not seeing locally what the GHA is reporting.

I fixed the issues in the last push-force. In a separate PR I will bring the `ruff` formater changes to the whole code (only a few changes are reported).

oesteban:
Outstanding work - thanks!

PR #147, opened 2024-03-28
STYLE: Format the code base style

jhlegarreta:
Format the code base style.

Add the corresponding command to the contribution checks GitHub Actions workflow so that the differences that need to be applied are actually shown on GitHub.


jhlegarreta:
Re: https://github.com/nipreps/eddymotion/pull/144#issuecomment-2025509448

jhlegarreta:
@arokem @oesteban @effigies it is probably worthwhile considering merging to avoid conflicts should other commits get into the way.

Issue #111, opened 2022-12-09
LOGO split - strategies other than random to traverse all orientations


oesteban:
I believe I have discussed this with @tmspvn, but I might be wrong, and it may have been with @arokem.

Either way, it could be beneficial (esp. with sophisticated models) to first align the lowest b-values and then go for the higher ones. This way, the orientations of lowest SNR are left to the last, and in LOGO splits the rest of the dataset is better aligned.

PR #149, opened 2024-03-30
ENH: Simplify `eddymotion.estimator.EddyMotionEstimator.fit`

### LINKED ISSUE #142: C901 Exception Removal
Remove the C901 exception from the Ruff settings in pyproject.toml and resolve the error occurring in estimator.py within src/eddymotion, as detailed in PR #139

jhlegarreta:
Simplify `eddymotion.estimator.EddyMotionEstimator.fit`: break down different parts into separate methods.

Fixes:
```
src/eddymotion/estimator.py:43:9: C901 `fit` is too complex (23 > 10)
```

raised by `ruff`.

Remove the `C901` error exception rule from the `ruff` linter whitelist.

Fixes #142.

### DIFF
--- a/src/eddymotion/estimator.py
+++ b/src/eddymotion/estimator.py
@@ -21,6 +21,7 @@
 #     https://www.nipreps.org/community/licensing/
 #
 """A model-based algorithm for the realignment of dMRI data."""
+
 from collections import namedtuple
 from pathlib import Path
 from tempfile import TemporaryDirectory, mkstemp
@@ -75,9 +76,7 @@ class EddyMotionEstimator:
             Number of parallel jobs.
         seed : :obj:`int` or :obj:`bool`
             Seed the random number generator (necessary when we want deterministic
-            estimation). If an integer, the value is used to initialize the
-            generator; if ``True``, the arbitrary value of ``20210324`` is used
-            to initialize it.
+            estimation). See :func:`_sort_dwdata_indices`.
         Return
         ------
         affines : :obj:`list` of :obj:`numpy.ndarray`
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Did not make these class members because the class does not have instance variables, and it is suggested them to be `static`, which does not make sense to me.

Some of these may better fit into a `utils.py` module, but that is beyond the scope of this PR/can be done in a separate PR as the need becomes more apparent.

jhlegarreta:
Doc build failure is unrelated:
https://github.com/nipreps/eddymotion/actions/runs/8493247406/job/23267121635?pr=149#step:5:488

jhlegarreta:
@arokem @oesteban @effigies it is probably worthwhile considering merging to avoid conflicts should other commits get into the way.

src/eddymotion/estimator.py
```diff
@@ -75,23 +76,18 @@ def fit(
...
         """
-        align_kwargs = align_kwargs or {}
 
-        _seed = None
-        if seed or seed == 0:
-            _seed = 20210324 if seed is True else seed
+        align_kwargs = align_kwargs or {}
 
-        rng = np.random.default_rng(_seed)
+        index_order = _sort_dwdata_indices(seed, len(dwdata))
```
oesteban, line 90: This is way better than it was - thanks. For the future, I think we should make this a generator that allows a number of ordering strategies (#111)

oesteban (approved):

src/eddymotion/estimator.py
jhlegarreta: :+1: IMO the strategies should be put into a `utils.py` module, as they may be of interest outside the estimator. Would also allow the estimator class/module become less complex/more readable.

oesteban: Totally agree - let's follow up on #111. Can you quickly draft your thoughts there?

jhlegarreta: Will do.

PR #150, opened 2024-04-02
ENH: Indexes sorting options

### LINKED ISSUE #111: LOGO split - strategies other than random to traverse all orientations

esavary:
Introduce iterators to rearrange the indexes used in logo_split. Sorting options include linear, random, in ascending order following gradient magnitude, and central symmetric.

closes #111

### DIFF
--- a/src/eddymotion/utils.py
+++ b/src/eddymotion/utils.py
@@ -127,7 +127,7 @@ def bvalue_iterator(size: int = None, **kwargs) -> Iterator[int]:
     """
     bvals = kwargs.get("bvals", None)
     if bvals is None:
-        raise TypeError('Keyword argument bvals is required')
+        raise TypeError("Keyword argument bvals is required")
     indexed_bvals = sorted([(round(b, 2), i) for i, b in enumerate(bvals)])
     return (index[1] for index in indexed_bvals)
 


src/eddymotion/estimator.py
```diff
@@ -75,16 +77,69 @@ def fit(
...
         ------
         affines : :obj:`list` of :obj:`numpy.ndarray`
             A list of :math:`4 \times 4` affine matrices encoding the estimated
             parameters of the deformations caused by head-motion and eddy-currents.
+        index_order : :obj:`numpy.ndarray`
+            The shuffled index order of the gradients. Only returned if `testing` is True.
 
         """
+
+        def handle_gradient_ordering():
```
jhlegarreta, line 96: Re: https://github.com/nipreps/eddymotion/pull/149#discussion_r1549864483

All the methods here do not access any instance variable, and any IDE will suggest them becoming static, as they were prior to 8153649. However the `actions` dictionary was triggering errors because it was taking static methods. All this could be avoided if these methods were refactored into a `utils.py` (or similar) module that takes care of sorting the DWI volumes. In fact, (i) the estimator should not take care of this at all: it should receive the DWI volumes already sorted, and (ii) these methods may be useful elsewhere.

Related to this, the strategies could be put into an `enum` so that it becomes clear which are the sorting possibilities, and they are documented and available at a single place. However, I know that using `enum`s is not very popular across many major scientific Python packages.

esavary: Sounds like a great idea! I agree, I'll externalize the functions into a new module, util.py. Thanks for the suggestion!

src/eddymotion/utils.py
```diff
@@ -0,0 +1,171 @@
...
+    _seed = None
+    if seed or seed == 0:
+        _seed = 20210324 if seed is True else seed
+
+    rng = np.random.default_rng(_seed)
+
+    index_order = np.arange(len(dwdata))
+    rng.shuffle(index_order)
+
+    return index_order
```
oesteban, line 125: Let's return just lists of integers (I'd prefer these were generators, but for a first stab looks good with a list):

```suggestion
    return index_order.to_list()
```

oesteban (commented):
Some initial comments

src/eddymotion/utils.py
```diff
@@ -0,0 +1,171 @@
...
+
+        for i in np.arange(1, middle_point+1):
+            index_order[2*i-1] = old_index[middle_point + i]
+            index_order[2*i] = old_index[middle_point - i]
+    else:
+        middle_point = int(len(old_index) / 2)
+        index_order[0] = old_index[middle_point]
+        for i in np.arange(1, middle_point+1):
+            index_order[2*i-1] = old_index[middle_point + i]
+            index_order[2*i] = old_index[middle_point - i]
```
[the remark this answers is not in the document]
esavary: It returns an IndexError for odd numbers, but I'll try to find a similar solution.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,176 @@
...
+            index_order[2 * i - 1] = old_index[middle_point + i]
+            index_order[2 * i] = old_index[middle_point - i]
+    else:
+        middle_point = int(len(old_index) / 2)
+        index_order[0] = old_index[middle_point]
+        for i in np.arange(1, middle_point + 1):
+            index_order[2 * i - 1] = old_index[middle_point + i]
+            index_order[2 * i] = old_index[middle_point - i]
+
+    return index_order
```
oesteban, line 176:
```suggestion
    linear = list(range(size))
    half1, half2 = list(reversed(linear[:size // 2])), linear[size // 2:]
    index_order = [
        sub[item] for item in range(len(half1))
        for sub in [half1, half2]
    ]
    if size % 2:  # If size is odd number, append last element
        index_order.append(half2[-1])
    return index_order
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+    Sort the DWI data volume indices in a central symmetric manner.
+
+    Parameters:
+    dwdata : :obj:`~eddymotion.dmri.DWI`
+        DWI dataset, represented by this tool's internal type.
+
+    Returns:
+    numpy.ndarray: The sorted index order.
+
```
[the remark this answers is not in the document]
oesteban: Added a test for odd number `size` above.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,176 @@
...
+    Returns
+    -------
+    numpy.ndarray: The sorted index order.
+    """
+    last_column = dwdata.gradients[:, -1]
+    index_order = np.argsort(last_column)
+    return index_order
+
+
+def centralsym_action(dwdata):
```
oesteban, line 146:
```suggestion
def centralsym_action(size, **kwargs):
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,176 @@
...
+        return linear_action(dwdata)
+    elif strategy == SortingStrategy.RANDOM:
+        return random_action(dwdata, seed)
+    elif strategy == SortingStrategy.BVALUE:
+        return bvalue_action(dwdata)
+    elif strategy == SortingStrategy.CENTRALSYM:
+        return centralsym_action(dwdata)
+    else:
+        raise ValueError("Invalid sorting strategy")
+
```
oesteban, line 80: Because of later commits my comment has gone away. As we discussed, let's leave this for a future PR.
```suggestion
```

test/test_splitting.py
oesteban, line 1: these changes are stylistic and should be resolved on a separate PR addressing #151.

test/test_model.py
oesteban, line 1: stylistic, please resolve in a separate PR addressing #151

src/eddymotion/utils.py
```diff
@@ -0,0 +1,176 @@
...
+    Sort the DWI data volume indices linearly
+
+    Parameters
+    ----------
+    dwdata : :obj:`~eddymotion.dmri.DWI` DWI dataset, represented by this tool's internal type.
+
+    Returns
+    -------
+    index_order : :obj:`numpy.ndarray`
+    The sorted index order.
```
oesteban, line 93: Comment lost after new commits:
```suggestion
    Parameters
    ----------
    size : :obj:`int`
        Number of volumes in the dataset
        (for instance, the number of orientations in a DWI)

    Returns
    -------
    :obj:`list` of :obj:`int`
        The sorted index order.

    Examples
    --------
    >>> linear_action(10)
    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,176 @@
...
+    Sort the DWI data volume indices in a central symmetric manner.
+
+    Parameters
+    ----------
+    dwdata : :obj:`~eddymotion.dmri.DWI` DWI dataset, represented by this tool's internal type.
+
+    Returns
+    -------
+    numpy.ndarray: The sorted index order.
+
```
oesteban, line 157:
```suggestion
    Parameters
    ----------
    size : :obj:`int`
        Number of volumes in the dataset
        (for instance, the number of orientations in a DWI)

    Returns
    -------
    :obj:`list` of :obj:`int`
        The sorted index order.
```

src/eddymotion/utils.py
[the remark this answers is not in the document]
oesteban: Right - resolved in https://github.com/nipreps/eddymotion/pull/150#discussion_r1552481306 and updated my suggestion of doctest to test the odd number size too.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,171 @@
...
+    Returns:
+    index_order : :obj:`numpy.ndarray`
+    The sorted index order.
+    """
+    index_order = np.arange(len(dwdata))
+
+    return index_order
+
+
+def random_action(dwdata, seed=None):
```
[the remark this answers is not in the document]
esavary: I couldn't apply this suggestion directly, but I'll introduce it in my next commit.

src/eddymotion/utils.py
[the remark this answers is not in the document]
esavary: I'll include this in my next commit

src/eddymotion/utils.py
esavary: I don't think it's working for odd numbers, e.g., for the sequence [0, 1, 2, 3, 4, 5, 6], it returns [2, 3, 1, 4, 0, 5, 6]. From our discussion yesterday, I was expecting either [3, 4, 2, 5, 1, 6, 0] or [3, 2, 4, 1, 5, 0, 6], but perhaps I misunderstood. Could you confirm whether we should expect [3, 4, 2, 5, 1, 6, 0] or [3, 2, 4, 1, 5, 0, 6]?

test/test_splitting.py
esavary: Yes, sure, I'll revert the changes for those files. Regarding future PRs, what are the guidelines for making small stylistic changes? Should we always correct them in a new PR? I'm asking because of the discussion in https://github.com/nipreps/eddymotion/pull/139#issuecomment-2015229270. No problem either way, just seeking clarification.

src/eddymotion/utils.py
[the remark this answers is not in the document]
oesteban: Why not committing the suggestion?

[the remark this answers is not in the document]
esavary: So I'll add doctests to the other methods at the same time. (And I recall I have to do it while the comment is open).

src/eddymotion/utils.py
esavary: Yes sure. I agree with the expected solution, and I committed the suggestion to confirm there is an issue with the implementation of the method.

src/eddymotion/estimator.py
oesteban, line 1: As we agreed in the meeting, let's not change this yet. Let's first have a battery of indexers available and in a future PR we connect that work to the Estimator.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,150 @@
...
+    - CENTRALSYM: Sorts the items based on their central symmetry.
+    """
+
+    LINEAR = "linear"
+    RANDOM = "random"
+    BVALUE = "bvalue"
+    CENTRALSYM = "centralsym"
+
+
+
```
oesteban, line 47: Let's work on this on a following-up PR later.

```suggestion
```

src/eddymotion/utils.py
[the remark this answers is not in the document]
oesteban: I don't understand why it cannot be applied directly, but sure.

src/eddymotion/estimator.py
esavary: Sure, I reverted the changes in 4998afa

src/eddymotion/utils.py
```diff
@@ -0,0 +1,150 @@
...
+    Returns
+    -------
+    :obj:`list` of :obj:`int`
+        The sorted index order.
+    """
+    linear = list(range(size))
+    half1, half2 = list(reversed(linear[:size // 2])), linear[size // 2:]
+    index_order = [
+        sub[item] for item in range(len(half1))
+        for sub in [half1, half2]
```
oesteban, line 128: Okay, I think I caught the problem you were hinting at with odd numbers:

```suggestion
        for sub in [half2, half1]
```

With odd numbers, `half2` is one element longer than `half1`, so starting with an element in `half2` makes more sense. This will require updating the two tests above (added within this review).

src/eddymotion/utils.py
```diff
@@ -0,0 +1,150 @@
...
+    Parameters
+    ----------
+    size : :obj:`int`
+        Number of volumes in the dataset
+        (for instance, the number of orientations in a DWI)
+
+    Examples
+    --------
+    >>> centralsym_action(10)
+    [4, 5, 3, 6, 2, 7, 1, 8, 0, 9]
```
oesteban, line 115:
```suggestion
    [5, 4, 6, 3, 7, 2, 8, 1, 9, 0]
```

src/eddymotion/utils.py
esavary: > 

Before we merge, there's still this issue with the implementation of the method. Would you be ok with me making modifications to your code, or would you prefer to handle the changes yourself?

src/eddymotion/utils.py
```diff
@@ -0,0 +1,150 @@
...
+    >>> centralsym_action(11)
+    [4, 5, 3, 6, 2, 7, 1, 8, 0, 9, 10]
```
oesteban, line 117:
```suggestion
    [5, 4, 6, 3, 7, 2, 8, 1, 9, 0, 10]
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Examples
+    --------
+    >>> linear_action(10)
+    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+    """
+    index_order = np.arange(len(dwdata))
+
+    return index_order
```
oesteban, line 53: IMHO all these would be more elegant if they returned generators:

```suggestion
    >>> list(linear_action(10))
    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

    """
    return range(size)
```

src/eddymotion/utils.py
```diff
@@ -86,7 +84,7 @@ def bvalue_action(gradients):
 
     Parameters
     ----------
-    dwdata : :obj:`~eddymotion.dmri.DWI` DWI dataset, represented by this tool's internal type.
+    gradients : :obj:`numpy.ndarray` gradient table of the dataset
```
oesteban, line 85: This doesn't need the full table (b-vectors + b-values), here just the b-values should be used:

```suggestion
    bvalues : :obj:`list`
        List of b-values corresponding to all orientations of the dataset.
```

This makes me realize we want `size` to be also a `kwarg` because for this iterator we do not need the size (it is implicitly the same as the length of the b-values list).

src/eddymotion/utils.py
```diff
@@ -0,0 +1,133 @@
...
+#     https://www.nipreps.org/community/licensing/
+#
+"""Utils to sort the DWI data volume indices"""
+
+from enum import Enum
+
+import numpy as np
+
+
+def linear_action(size, **kwargs):
```
oesteban, line 28: If we want all the iterators to be called exactly the same (and we want), then size should be optional for all iterators:

```suggestion
def linear_action(size=None, **kwargs):
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,133 @@
...
+    :obj:`list` of :obj:`int`
+        The sorted index order.
+
+    Examples
+    --------
+    >>> list(linear_action(10))
+    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+    """
+    return range(size)
```
oesteban, line 49: And now, we can make it smarter and calculate size from bvals if they are given instead of size:

```suggestion
    if size is None and 'bvals' in kwargs:
        size = len(kwargs['bvals'])
    if size is None:
       raise TypeError("Cannot build iterator without size") 
    return range(size)
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def random_action(size, **kwargs):
```
oesteban, line 52:
```suggestion
def random_action(size=None, **kwargs):
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,133 @@
...
+
+    rng = np.random.default_rng(_seed)
+
+    index_order = np.arange(size)
+    rng.shuffle(index_order)
+
+    return index_order.to_list()
+
+
+def bvalue_action(gradients):
```
oesteban, line 79:
```suggestion
def bvalue_action(size=None, **kwargs):
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,133 @@
...
+    Examples
+    --------
+    >>> bvalue_action(np.array([[1, 1, 1, 1], [4, 4, 4, 4], [2, 2, 2, 2]]))
+    array([0, 2, 1])
+
+    Returns
+    -------
+    numpy.ndarray: The sorted index order.
+    """
+    last_column = gradients[:, -1]
```
oesteban, line 96:
```suggestion
    bvals = kwargs.get('bvals', None)
    if bvals is None:
        raise TypeError('Keyword argument bvals is required')
    indexed_bvals = sorted([(round(b, 2), i) for i, b in enumerate(bvals)])
    return [index[1] for index in indexed_bvals]
    
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,133 @@
...
+    index_order = np.argsort(last_column)
+    return index_order
+
+
+def centralsym_action(size, **kwargs):
```
oesteban, line 101:
```suggestion
def centralsym_action(size=None, **kwargs):
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,133 @@
...
+    [5, 4, 6, 3, 7, 2, 8, 1, 9, 0]
+    >>> centralsym_action(11)
+    [5, 4, 6, 3, 7, 2, 8, 1, 9, 0, 10]
+
+    Returns
+    -------
+    :obj:`list` of :obj:`int`
+        The sorted index order.
+    """
+    linear = list(range(size))
```
oesteban, line 123:
```suggestion
    if size is None and 'bvals' in kwargs:
        size = len(kwargs['bvals'])
    if size is None:
       raise TypeError("Cannot build iterator without size") 
    linear = list(range(size))
```

oesteban:
> After the build, when inspecting './docs/api/eddymotion.rst', however, I observe that 'eddymotion.utils' is included in the toctree. Any insights on what could be causing this issue

To generate the documentation for developers we use sphinx and a plugin called autodoc. Autodoc goes through all source code files and documents stuff based on your docstrings.

In this case, the error is telling you that autodoc generated a documentation file for the new module at `eddymotion/docs/api/eddymotion.utils.rst`, but that file (a restructured text formatted file, .rst) is not then linked anywhere - so the document is orphaned.

You want to add this document in a suitable place of the docs:

https://github.com/nipreps/eddymotion/blob/a1bcacd6117a9f9ce95424dae7fb17c893a885a1/docs/developers.rst#L24-L35

should now be:

```
Library API (application program interface)
-------------------------------------------
Information on specific functions, classes, and methods.

.. toctree::
   :glob:

   api/eddymotion.data
   api/eddymotion.data.dmri
   api/eddymotion.estimator
   api/eddymotion.model
   api/eddymotion.utils
   api/eddymotion.viz
```

esavary:
test_estimator.py is currently failing in the CircleCI check routine. However, the changes in this PR should not have affected this, as the files modified in this PR do not impact the code execution. Additionally, some of the commits in this PR passed the test previously. Should I open an issue about this?

esavary:
> > After the build, when inspecting './docs/api/eddymotion.rst', however, I observe that 'eddymotion.utils' is included in the toctree. Any insights on what could be causing this issue
> 
> To generate the documentation for developers we use sphinx and a plugin called autodoc. Autodoc goes through all source code files and documents stuff based on your docstrings.
> 
> In this case, the error is telling you that autodoc generated a documentation file for the new module at `eddymotion/docs/api/eddymotion.utils.rst`, but that file (a restructured text formatted file, .rst) is not then linked anywhere - so the document is orphaned.
> 
> You want to add this document in a suitable place of the docs:
> 
> https://github.com/nipreps/eddymotion/blob/a1bcacd6117a9f9ce95424dae7fb17c893a885a1/docs/developers.rst#L24-L35
> 
> should now be:
> 
> ```
> Library API (application program interface)
> -------------------------------------------
> Information on specific functions, classes, and methods.
> 
> .. toctree::
>    :glob:
> 
>    api/eddymotion.data
>    api/eddymotion.data.dmri
>    api/eddymotion.estimator
>    api/eddymotion.model
>    api/eddymotion.utils
>    api/eddymotion.viz
> ```

Thanks a lot, I changed this in 3c0b36d.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Returns
+    -------
+    :obj:`list` of :obj:`int`
+        The sorted index order.
+    """
+    bvals = kwargs.get('bvals', None)
+    if bvals is None:
+        raise TypeError('Keyword argument bvals is required')
+    indexed_bvals = sorted([(round(sum(sublist), 2), i) for i, sublist in enumerate(bvals)])
```
esavary, line 111: I changed it because it was causing a TypeError for a bvals that is a list. Feel free to suggest a more elegant solution.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+
+    if size is None and 'bvals' in kwargs:
+        size = len(kwargs['bvals'])
+    if size is None:
+        raise TypeError("Cannot build iterator without size")
+
+    _seed = kwargs.get('seed', None)
+    if kwargs.get('seed', None) or kwargs.get('seed', None) == 0:
+        _seed = 20210324 if kwargs.get('seed', None) is True else kwargs.get('seed', None)
```
oesteban, line 79: I don't get this (??)

src/eddymotion/utils.py
```diff
@@ -0,0 +1,149 @@
...
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Utils to sort the DWI data volume indices"""
+
+import numpy as np
+
+
+def linear_action(size=None, **kwargs):
```
oesteban, line 28: What do you mean by "action"? I think we just want "iterator" here (`linear_iterator` and so on so forth).

src/eddymotion/utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Parameters
+    ----------
+    size : :obj:`int`
+        Number of volumes in the dataset
+        (for instance, the number of orientations in a DWI)
+
+    Returns
+    -------
+    :obj:`list` of :obj:`int`
```
oesteban, line 40: We can better document this:

```suggestion
    :obj:`~typing.Generator`
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,149 @@
...
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Utils to sort the DWI data volume indices"""
+
+import numpy as np
```
oesteban, line 25: Let's simplify the implementation of the alternating from center iterator:

```suggestion
import numpy as np
from itertools import chain, zip_longest
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,149 @@
...
+    size : :obj:`int`
+        Number of volumes in the dataset
+        (for instance, the number of orientations in a DWI)
+
+    Examples
+    --------
+    >>> centralsym_action(10)
+    [5, 4, 6, 3, 7, 2, 8, 1, 9, 0]
+    >>> centralsym_action(11)
+    [5, 4, 6, 3, 7, 2, 8, 1, 9, 0, 10]
```
oesteban, line 136: If we turn this into a generator, then we need to expand lists before testing:
```suggestion
    >>> list(centralsym_iterator(10))
    [5, 4, 6, 3, 7, 2, 8, 1, 9, 0]
    >>> list(centralsym_iterator(11))
    [5, 4, 6, 3, 7, 2, 8, 1, 9, 0, 10]
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Returns
+    -------
+    :obj:`list` of :obj:`int`
+        The sorted index order.
+    """
```
oesteban:
```suggestion
    :obj:`~typing.Generator`
        The generator iterable with the appropriate order.

    """
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,149 @@
...
+        raise TypeError("Cannot build iterator without size")
+    linear = list(range(size))
+    half1, half2 = list(reversed(linear[:size // 2])), linear[size // 2:]
+    index_order = [
+        sub[item] for item in range(len(half1))
+        for sub in [half2, half1]
+    ]
+    if size % 2:  # If size is odd number, append last element
+        index_order.append(half2[-1])
+    return index_order
```
oesteban, line 151:
```suggestion
    if size is None and 'bvals' in kwargs:
        size = len(kwargs['bvals'])
    if size is None:
        raise TypeError("Cannot build iterator without size")
    linear = list(range(size))
    return (
        x for x in chain.from_iterable(zip_longest(
            linear[size // 2:],
            reversed(linear[:size // 2]),
        ))
        if x is not None
    )
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,151 @@
...
+    --------
+    >>> bvalue_action(bvals=[0.0, 0.0, 1000.0, 1000.0, 700.0, 700.0, 2000.0, 2000.0, 0.0])
+    [0, 1, 8, 4, 5, 2, 3, 6, 7]
+
+    """
+    bvals = kwargs.get('bvals', None)
+    if bvals is None:
+        raise TypeError('Keyword argument bvals is required')
+    indexed_bvals = sorted([(round(sum(sublist), 2), i) for i, sublist in enumerate(bvals)])
+    return [index[1] for index in indexed_bvals]
```
oesteban, line 113: Let's roll back and make this an iterator too:
```suggestion
    >>> list(bvalue_iterator(bvals=[0.0, 0.0, 1000.0, 1000.0, 700.0, 700.0, 2000.0, 2000.0, 0.0]))
    [0, 1, 8, 4, 5, 2, 3, 6, 7]

    """
    bvals = kwargs.get('bvals', None)
    if bvals is None:
        raise TypeError('Keyword argument bvals is required')
    indexed_bvals = sorted([(round(b, 2), i) for i, b in enumerate(bvals)])
    return (index[1] for index in indexed_bvals)
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,151 @@
...
+        (for instance, the number of orientations in a DWI)
+
+    Returns
+    -------
+    :obj:`~typing.Generator`
+        The sorted index order.
+
+    Examples
+    --------
+    >>> list(linear_action(10))
```
oesteban, line 45: Name should be updated here:
```suggestion
    >>> list(linear_iterator(10))
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,151 @@
...
+    Parameters
+    ----------
+    size : :obj:`int`
+        Number of volumes in the dataset
+        (for instance, the number of orientations in a DWI)
+
+    Returns
+    -------
+    :obj:`list` of :obj:`int`
+        The sorted index order.
```
oesteban, line 69: Examples? This would be interesting to test we are properly handling the seed.

oesteban (commented):
Shaping up - I'm suggesting to change all these to generators and simplifying `centralsym_iterator()`

src/eddymotion/utils.py
esavary: Added in [a43fdde](https://github.com/nipreps/eddymotion/pull/150/commits/a43fdde0d3aed8558106f482916847709c891dee)

src/eddymotion/utils.py
esavary: It's inspired from the _sort_dwdata_indices() function in estimator.py. I thought that in the future, we would want to externalize the initialization of the seed in the random generator instead of the backend function _sort_dwdata_indices() but feel free to remove if it's not the right place for this.

However, there seems to be a bug in the original function -- when the seed argument is set to 0, the condition "if seed or seed == 0:" considers zero as False, and the value of the seed is not set correctly to 20210324, but instead stays at 0. I am not sure if this bug is only present in some Python versions, but I am currently using version 3.11. Should I post an issue about this, or is it okay if I propose a fix this in this PR?

src/eddymotion/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Examples
+    --------
+    >>> random_iterator(10, seed=0, test_seed=True)
+    20210324
+    >>> random_iterator(10, seed=True, test_seed=True)
+    20210324
+    >>> random_iterator(10, seed=42, test_seed=True)
+    42
```
oesteban, line 81: What is the purpose of `test_seed`? Altering the function implementation to ease testing is backwards.

If the seed is not working, these tests will fail more often than they can possibly pass because the chances that the random sample matches the expected value of the test are really low.

```suggestion
    >>> random_iterator(15, seed=0)
    [12, 2, 7, 8, 14, 4, 9, 13, 6, 3, 5, 0, 1, 10, 11]
    >>> random_iterator(15, seed=True)
    [13, 2, 8, 12, 10, 4, 9, 3, 0, 1, 14, 6, 11, 7, 5]
    >>> random_iterator(15, seed=42)
    [0, 13, 8, 4, 7, 2, 3, 5, 10, 1, 12, 6, 11, 14, 9]
```

src/eddymotion/utils.py
oesteban: I think you are referring to this code added by @jhlegarreta:

https://github.com/nipreps/eddymotion/blob/a1bcacd6117a9f9ce95424dae7fb17c893a885a1/src/eddymotion/estimator.py#L261-L263

Here, I think the purpose is to set the default value (20210324) if (and only if) the seed argument is a `True` value, and when seed is the integer zero (`0`), then set the seed to zero (as it is an acceptable value).

Arguably, this code probably would have worked just fine if you do:

```
seed = 20210324 if seed is True else seed
```

That would not be fully equivalent, as `False` would remain `False` with my code. However, `False` is actually an acceptable value by the random number generator seed. I will send a suggestion with this (since this has become outdated, I cannot do it here).

src/eddymotion/utils.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+
+    if size is None and 'bvals' in kwargs:
+        size = len(kwargs['bvals'])
+    if size is None:
+        raise TypeError("Cannot build iterator without size")
+
+    _seed = kwargs.get('seed', None)
+    if _seed is True or _seed == 0:
+        _seed = 20210324 if kwargs.get('seed', None) in [True, 0] else kwargs.get('seed', None)
```
oesteban, line 92:
```suggestion
    _seed = 20210324 if _seed is True else _seed
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+    rng = np.random.default_rng(_seed)
+
+    index_order = np.arange(size)
+    rng.shuffle(index_order)
+
+    if kwargs.get('test_seed', False) is True:
+        return _seed
+
```
oesteban, line 101:
```suggestion
```

src/eddymotion/utils.py
oesteban: False is cast to zero -- I'm going to update my suggestion so that False is made None.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,166 @@
...
+    if size is None and 'bvals' in kwargs:
+        size = len(kwargs['bvals'])
+    if size is None:
+        raise TypeError("Cannot build iterator without size")
+
+    _seed = kwargs.get('seed', None)
+    if _seed is True or _seed == 0:
+        _seed = 20210324 if kwargs.get('seed', None) in [True, 0] else kwargs.get('seed', None)
+
+    rng = np.random.default_rng(_seed)
```
oesteban, line 94:
```suggestion
    rng = np.random.default_rng(
        None if _seed is False else _seed,
    )
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,163 @@
...
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Utils to sort the DWI data volume indices"""
+
+import numpy as np
```
oesteban, line 25: Let's stick with the standard library for this:
```suggestion
import random
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,163 @@
...
+        size = len(kwargs['bvals'])
+    if size is None:
+        raise TypeError("Cannot build iterator without size")
+
+    _seed = kwargs.get('seed', None)
+    _seed = 20210324 if _seed is True else _seed
+
+    rng = np.random.default_rng(
+        None if _seed is False else _seed,
+    )
```
oesteban, line 95:
```suggestion
    random.seed(None if _seed is False else _seed)
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+    index_order = np.arange(size)
+    rng.shuffle(index_order)
```
oesteban, line 98:
```suggestion
    index_order = list(range(size))
    random.shuffle(index_order)
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,163 @@
...
+    return index_order.tolist()
```
oesteban, line 99: Make a generator:
```suggestion
    return (x for x in index_order)
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,163 @@
...
+        If True, return the seed value used for the random number generator.
+
+    Returns
+    -------
+    :obj:`~typing.Generator`
+        The sorted index order.
+
+    Examples
+    --------
+    >>> random_iterator(15, seed=0)  # seed is 0
```
esavary, line 76: This test is failing on my computer locally. I'm getting a different value compared to yours @oesteban . Specifically, I'm getting [2, 11, 3, 10, 0, 4, 7, 5, 14, 12, 6, 9, 13, 8, 1]. Any idea why?

src/eddymotion/utils.py
```diff
@@ -0,0 +1,163 @@
...
+    [12, 2, 7, 8, 14, 4, 9, 13, 6, 3, 5, 0, 1, 10, 11]
+    >>> random_iterator(15, seed=True)  # seed is the default value 20210324
+    [13, 2, 8, 12, 10, 4, 9, 3, 0, 1, 14, 6, 11, 7, 5]
+    >>> random_iterator(15, seed=42)  # seed is 42
```
esavary, line 80: Same here, I obtain [6, 7, 9, 3, 0, 11, 14, 10, 5, 2, 4, 12, 1, 13, 8].

src/eddymotion/utils.py
```diff
@@ -0,0 +1,163 @@
...
+def random_iterator(size=None, **kwargs):
+    """Sort the DWI data volume indices.
+
+    Parameters
+    ----------
+    size : :obj:`int`
+        Number of volumes in the dataset
+        (for instance, the number of orientations in a DWI)
+    test_seed : :obj:`bool`
+        If True, return the seed value used for the random number generator.
```
oesteban, line 67:
```suggestion
    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None``
        If :obj:`int` or :obj:`str` or ``None``, initializes the seed of Python's random generator
        with the given value.
        If ``False``, the random generator is passed ``None``.
        If ``True``, a default seed value is set.
```

src/eddymotion/utils.py
esavary: Ok, thanks for clarifying. I wondered if zero was avoided for another reason.

src/eddymotion/utils.py
oesteban, line 80:
```suggestion
    >>> list(random_iterator(15, seed=0))  # seed is 0
    [12, 2, 7, 8, 14, 4, 9, 13, 6, 3, 5, 0, 1, 10, 11]
    >>>  # seed is True -> the default value 20210324 is set
    >>> list(random_iterator(15, seed=True))
    [13, 2, 8, 12, 10, 4, 9, 3, 0, 1, 14, 6, 11, 7, 5]
    >>> list(random_iterator(15, seed=20210324))
    [13, 2, 8, 12, 10, 4, 9, 3, 0, 1, 14, 6, 11, 7, 5]
    >>> list(random_iterator(15, seed=42))  # seed is 42
```

oesteban (commented):
Suggestions to avoid numpy for the time being. Also make the random range iterator a generator.

src/eddymotion/utils.py
oesteban: No idea - I might have copied the wrong order.

oesteban:
> Suggestions to avoid numpy for the time being. Also make the random range iterator a generator.

These for sure will make the test values invalid - we'll need to re-write them.

esavary:
> > Suggestions to avoid numpy for the time being. Also make the random range iterator a generator.
> 
> These for sure will make the test values invalid - we'll need to re-write them.

I changed them in [ed20ca8](https://github.com/nipreps/eddymotion/pull/150/commits/ed20ca8c98778f464934c8ce9739a48b4c28c8c8)

src/eddymotion/utils.py
```diff
@@ -0,0 +1,167 @@
...
+    bvals = kwargs.get('bvals', None)
+    if bvals is None:
+        raise TypeError('Keyword argument bvals is required')
+    indexed_bvals = sorted([(round(b, 2), i) for i, b in enumerate(bvals)])
+    return (index[1] for index in indexed_bvals)
+
+
+def centralsym_iterator(size=None, **kwargs):
+    """
+    Sort the DWI data volume indices in a central symmetric manner.
```
oesteban, line 135:
```suggestion
    Traverse the dataset starting from the center and alternatingly progressing to the sides.
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,167 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2022 The NiPreps Developers <nipreps@gmail.com>
```
oesteban, line 4:
```suggestion
# Copyright 2024 The NiPreps Developers <nipreps@gmail.com>
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,169 @@
...
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Iterators to traverse the volumes in a 4D image."""
+
+
+def linear_iterator(size: int = None, **kwargs) -> Generator[int, None, None]:
```
oesteban, line 30:
```suggestion
def linear_iterator(size: int = None, **kwargs) -> Generator[int]:
```

src/eddymotion/utils.py
```diff
@@ -22,23 +26,20 @@
...
 
-def linear_iterator(size=None, **kwargs):
+def linear_iterator(size: int = None, **kwargs) -> Generator[int, None, None]:
     """
     Traverse the dataset volumes in ascending order.
 
     Parameters
     ----------
-    size : :obj:`int`
+    size : int, optional
```
oesteban, line 36:
```suggestion
    size : :obj:`int`
```

src/eddymotion/utils.py
```diff
@@ -22,23 +26,20 @@
...
         Number of volumes in the dataset
         (for instance, the number of orientations in a DWI)
 
     Returns
     -------
-    :obj:`~typing.Generator`
+    Generator[int, None, None]
```
oesteban, line 42:
```suggestion
    :obj:`~typing.Generator`
```

src/eddymotion/utils.py
```diff
@@ -55,24 +56,24 @@ def linear_iterator(size=None, **kwargs):
     return range(size)
 
 
-def random_iterator(size=None, **kwargs):
+def random_iterator(size: int = None, **kwargs) -> Generator[int, None, None]:
```
oesteban, line 59:
```suggestion
def random_iterator(size: int = None, **kwargs) -> Generator[int]:
```

src/eddymotion/utils.py
```diff
@@ -55,24 +56,24 @@ def linear_iterator(size=None, **kwargs):
...
     """
     Traverse the dataset volumes randomly.
 
     Parameters
     ----------
-    size : :obj:`int`
+    size : int, optional
```
oesteban, line 65:
```suggestion
    size : :obj:`int`
```

src/eddymotion/utils.py
```diff
@@ -55,24 +56,24 @@ def linear_iterator(size=None, **kwargs):
...
         Number of volumes in the dataset
         (for instance, the number of orientations in a DWI)
-    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None``
-        If :obj:`int` or :obj:`str` or ``None``, initializes the seed of Python's random generator
+    seed : int or bool or None, optional
+        If int or str or None, initializes the seed of Python's random generator
```
oesteban, line 69:
```suggestion
    seed : :obj:`int` or :obj:`bool` or :obj:`bool` or ``None``
        If :obj:`int` or :obj:`str` or ``None``, initializes the seed of Python's random generator
```

src/eddymotion/utils.py
```diff
@@ -55,24 +56,24 @@ def linear_iterator(size=None, **kwargs):
...
         with the given value.
-        If ``False``, the random generator is passed ``None``.
-        If ``True``, a default seed value is set.
+        If False, the random generator is passed None.
+        If True, a default seed value is set.
```
oesteban, line 72:
```suggestion
        If ``False``, the random generator is passed ``None``.
        If ``True``, a default seed value is set.
```

src/eddymotion/utils.py
```diff
@@ -55,24 +56,24 @@ def linear_iterator(size=None, **kwargs):
...
 
     Returns
     -------
-    :obj:`~typing.Generator`
+    Generator[int, None, None]
```
oesteban, line 76:
```suggestion
    :obj:`~typing.Generator`
```

src/eddymotion/utils.py
```diff
@@ -104,18 +105,18 @@ def random_iterator(size=None, **kwargs):
     return (x for x in index_order)
 
 
-def bvalue_iterator(size=None, **kwargs):
+def bvalue_iterator(size: int = None, **kwargs) -> Generator[int, None, None]:
```
oesteban, line 108:
```suggestion
def bvalue_iterator(size: int = None, **kwargs) -> Generator[int]:
```

src/eddymotion/utils.py
```diff
@@ -104,18 +105,18 @@ def random_iterator(size=None, **kwargs):
...
     """
     Traverse the volumes in a DWI dataset by growing b-value.
 
     Parameters
     ----------
-    bvalues : :obj:`list`
+    bvals : list
```
oesteban, line 114:
```suggestion
    bvalues : :obj:`list`
```

src/eddymotion/utils.py
```diff
@@ -104,18 +105,18 @@ def random_iterator(size=None, **kwargs):
...
         List of b-values corresponding to all orientations of the dataset.
 
     Returns
     -------
-    :obj:`~typing.Generator`
+    Generator[int, None, None]
```
oesteban, line 119:
```suggestion
    :obj:`~typing.Generator`
```

src/eddymotion/utils.py
```diff
@@ -131,19 +132,19 @@ def bvalue_iterator(size=None, **kwargs):
...
 
-def centralsym_iterator(size=None, **kwargs):
+def centralsym_iterator(size: int = None, **kwargs) -> Generator[int, None, None]:
     """
     Traverse the dataset starting from the center and alternatingly progressing to the sides.
 
     Parameters
     ----------
-    size : :obj:`int`
+    size : int
```
oesteban, line 141:
```suggestion
    size : :obj:`int`
```

src/eddymotion/utils.py
```diff
@@ -131,19 +132,19 @@ def bvalue_iterator(size=None, **kwargs):
     return (index[1] for index in indexed_bvals)
 
 
-def centralsym_iterator(size=None, **kwargs):
+def centralsym_iterator(size: int = None, **kwargs) -> Generator[int, None, None]:
```
oesteban, line 135:
```suggestion
def centralsym_iterator(size: int = None, **kwargs) -> Generator[int]:
```

src/eddymotion/utils.py
```diff
@@ -131,19 +132,19 @@ def bvalue_iterator(size=None, **kwargs):
...
     ----------
-    size : :obj:`int`
+    size : int
         Number of volumes in the dataset
         (for instance, the number of orientations in a DWI)
 
     Returns
     -------
-    :obj:`~typing.Generator`
+    Generator[int, None, None]
```
oesteban, line 147:
```suggestion
    :obj:`~typing.Generator`
```

src/eddymotion/utils.py
```diff
@@ -22,23 +26,20 @@
 #
 """Iterators to traverse the volumes in a 4D image."""
 
-import random
-from itertools import chain, zip_longest
-
```
oesteban: Revert the removal of imports and add the Generation import here:

```suggestion
import random
from itertools import chain, zip_longest
from typing import Generator
```

src/eddymotion/utils.py
```diff
@@ -1,3 +1,7 @@
+import random
+from itertools import chain, zip_longest
+from typing import Generator
+
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
```
oesteban, line 5: Please, be careful, why these four lines here before the header?
```suggestion
# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
```

oesteban (commented):
The assistance tool for adding type annotations has introduced some problems (moved the imports to the top, before the license banner) and modified some docstrings that were better before.

src/eddymotion/utils.py
```diff
@@ -0,0 +1,169 @@
...
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Iterators to traverse the volumes in a 4D image."""
+
+
```
esavary, line 29: With Generator[int] instead of Generator[int,None,None] it returns this error "TypeError: Too few arguments for typing.Generator; actual 1, expected 3"

oesteban: You're right, as per the documentation: https://docs.python.org/3/library/typing.html#typing.Generator.

But, as the documentation also indicates at the bottom, we can instead annotate these as `Iterator[int]`

src/eddymotion/utils.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    if size is None and 'bvals' in kwargs:
+        size = len(kwargs['bvals'])
+    if size is None:
+        raise TypeError("Cannot build iterator without size")
+
+    return range(size)
+
+
+def random_iterator(size: int = None, **kwargs) -> Generator[int]:
```
oesteban, line 58:
```suggestion
def random_iterator(size: int = None, **kwargs) -> Iterator[int]:
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,169 @@
...
+    _seed = 20210324 if _seed is True else _seed
+
+    random.seed(None if _seed is False else _seed)
+
+    index_order = list(range(size))
+    random.shuffle(index_order)
+    return (x for x in index_order)
+
+
+def bvalue_iterator(size: int = None, **kwargs) -> Generator[int]:
```
oesteban, line 107:
```suggestion
def bvalue_iterator(size: int = None, **kwargs) -> Iterator[int]:
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,169 @@
...
+
+    """
+    bvals = kwargs.get('bvals', None)
+    if bvals is None:
+        raise TypeError('Keyword argument bvals is required')
+    indexed_bvals = sorted([(round(b, 2), i) for i, b in enumerate(bvals)])
+    return (index[1] for index in indexed_bvals)
+
+
+def centralsym_iterator(size: int = None, **kwargs) -> Generator[int]:
```
oesteban, line 134:
```suggestion
def centralsym_iterator(size: int = None, **kwargs) -> Iterator[int]:
```

src/eddymotion/utils.py
```diff
@@ -0,0 +1,168 @@
...
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Iterators to traverse the volumes in a 4D image."""
+
+import random
+from itertools import chain, zip_longest
+from typing import Generator
```
oesteban, line 27: Comment got outdated
```suggestion
from typing import Iterator
```

src/eddymotion/utils.py
oesteban: (`Iterator[int]` seems more informative to me)

oesteban:
Let's re-sync with main:

1. Check you have the main repo as a remote:
    ```
    git remote -v
    ```
    And if not, add it:
    ```
    git remote add upstream git@github.com:nipreps/eddymotion.git
    ```
2. Fetch the latest:
    ```
    git fetch upstream
    ```
3. Merge main into this branch:
    ```
    git merge upstream/main
    ```

oesteban (approved):

PR #158, opened 2024-04-12
ENH: Drop support for Python 3.8 and 3.9

### LINKED ISSUE #140: Limit Python versions that are supported?
Does nipreps have a policy about supported versions? Looks like currently several versions of Python are supported:

https://github.com/nipreps/eddymotion/blob/62a213a54ac2ac0e5211e7873a199c12a1751e19/.github/workflows/pythonpackage.yml#L19

Should we adopt [spec-0000](https://scientific-python.org/specs/spec-0000/)? That would imply dropping 3.8 and 3.9.

jhlegarreta:
Drop support for Python 3.8 and 3.9: require Python 3.10 or greater in line with Scientific Python SPEC 0:
https://scientific-python.org/specs/spec-0000/

Remove them from the CI matrices.

Use CircleCI's docker image so that the pacakge gets tested on Python 3.10:
https://hub.docker.com/u/cimg

Resolves #140.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
           paths:
             - /tmp/ants
 
-      - run: 
+      - run:
           name: Configure git (pacify datalad)
           command: |
             git config --global user.name "First Last"
@@ -55,7 +55,7 @@ jobs:
             - data-v1-{{ .Branch }}-
             - data-v1-main-
             - data-v1-
-      - run: 
+      - run:
           name: Pull down test data
           command: |
             if [[ ! -d "${TEST_DATA_HOME}" ]]; then
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           paths:
             - /tmp/ants
 
-      - run:
```
jhlegarreta, line 45: Had to make these changes to make ruff happy.

jhlegarreta:
Do the dependencies need to be updated?

jhlegarreta:
Ruff failure points to `.maint/paper_author_list.py`: unrelated:
https://github.com/nipreps/eddymotion/actions/runs/8655158166/job/23733655248?pr=158#step:4:14

jhlegarreta:
Re https://github.com/nipreps/eddymotion/pull/158#issuecomment-2050771198 fixed in PR #163.

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ orbs:
 jobs:
   tests:
     docker: # executor type
-      - image: nipreps/miniconda:py39_2205.0
+      - image: cimg/python:3.10
```
oesteban, line 8: Our miniconda image is mislabeled; it ships with 3.11 now (I will generate an appropriate label ASAP).
```suggestion
      - image: nipreps/miniconda:py39_2205.0
```

jhlegarreta: OK. Push forced in 0e46c3b.

Issue #155, opened 2024-04-09
test_ANTs_config_b0 failing on and off


jhlegarreta:
### What happened?

`eddymotion/test/test_estimator.py::test_ANTs_config_b0[1.0-0.0-1.0-0.3-0.1-0.3]` is failing on and off, e.g.
https://app.circleci.com/pipelines/github/nipreps/eddymotion/792/workflows/58453843-3f9d-4960-b1b7-387770345d7e/jobs/780?invite=true#step-110-54223_59
and a more recent run
https://app.circleci.com/pipelines/github/nipreps/eddymotion/941/workflows/71394aa4-00b8-40c9-b3c1-630d9edd1233/jobs/915

### What command did you use?

```shell
None: CircleCI is doing the job.
```

### What version of the software are you running?

main

### How are you running this software?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
coords = xfm.reference.ndcoords.T
        rms = np.sqrt(((xfm.map(coords) - xform.map(coords)) ** 2).sum(1)).mean()
>       assert rms < 0.8
E       assert 146.2325058746723 < 0.8

../src/eddymotion/test/test_estimator.py:79: AssertionError
```

### Additional information / screenshots

This is only being caught by CircleCI (and not on all runs) because the GitHub actions are not running the tests.

oesteban:
> This is only being caught by CircleCI (and not on all runs) because the GitHub actions are not running the tests.

Correct - the idea is that some tests ought to be checked visually, and the interface of CircleCI is way more amicable to quickly *see* the results.

Agreed we want to figure out this flaky test and give it a good solution :)

jhlegarreta:
CircleCI shows no errors since merging PR #160:
https://app.circleci.com/pipelines/github/nipreps/eddymotion?branch=main

So closing this.

PR #160, opened 2024-04-12
FIX: Set random seed in registration test

jhlegarreta:
Fix random seed in registration test for the sake of reproducibility.


jhlegarreta:
Not sure whether this would fix #155.

oesteban (approved):

PR #166, opened 2024-04-13
ENH: Check if `fit` has been called prior to `predict`

### LINKED ISSUE #113: Track model fit formally with an object member variable
Do not allow predicting if `model.is_fit_` is not `True` (credit for the idea to @arokem https://github.com/nipreps/eddymotion/pull/112#discussion_r1045142128)

jhlegarreta:
Check if `fit` has been called prior to `predict`: raise an exception if not.

Resolves #113.


jhlegarreta:
For now, **only added it for `TrivialB0Model` to first see how this fits into what was expected in the issue.**
 
A few notes:
- Following https://github.com/nipreps/eddymotion/pull/112#discussion_r1045142128, almost all of this has been borrowed from scikit-learn. If we keep this approach, maybe a mention to this will need to be made.
- In `scikit-learn`:
  - Their [base estimator](https://github.com/scikit-learn/scikit-learn/blob/872124551/sklearn/base.py#L152) does not have any `is_fitted_` variable. It does not have a `fit`/`predict` either; they are left for child classes.
  - They rely on a list of attributes (`coef_`, `estimator_`) existing for the estimator and provided as a parameter https://github.com/scikit-learn/scikit-learn/blob/872124551/sklearn/utils/validation.py#L1551. If none is provided, the estimator is considered fitted if there exists an attribute that ends with an underscore and does not start with a double underscore (e.g. could be the `is_fitted_` attribute).
- Their `__sklearn_is_fitted__` is defined in child classes, and is used as an alternative to the above two possibilities to check whether an estimator is fit.
- For now, I've left the new modules in the `eddymotion` package, but they should probably be moved to an e.g. `utils` modules.

Useful references:
https://scikit-learn.org/stable/developers/develop.html#developer-api-for-check-is-fitted
https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html#sphx-glr-auto-examples-developing-estimators-sklearn-is-fitted-py

Slightly related:
- eddymotion models (e.g. `TrivialB0Model`, etc.) do not inherit from `BaseModel`. Not sure if that is intended, but it is counter-intuitive or the rationale would need to be documented better.
- Adds to the confusion having an `EddyMotionEstimator` class that has a `fit` but not a `predict`. Maybe model vs. estimator naming should be reworked.

src/eddymotion/validation.py
```diff
@@ -0,0 +1,112 @@
...
+
+from eddymotion.exceptions import NotFittedError
+
+
+def _is_fitted(estimator, attributes=None, all_or_any=all):
+    """Determine if an estimator is fitted.
+
+    Parameters
+    ----------
+    estimator : estimator instance
```
jhlegarreta, line 34: This docstring needs to be changed once we figure out better the "model" vs "estimator" naming.

src/eddymotion/validation.py
```diff
@@ -0,0 +1,112 @@
...
+    fitted attributes (ending with a trailing underscore) and otherwise
+    raises a NotFittedError with the given message.
+
+    If an estimator does not set any attributes with a trailing underscore, it
+    can define a ``__eddymotion_is_fitted__`` method returning a boolean to
+    specify if the estimator is fitted or not.
+
+    Parameters
+    ----------
+    estimator : estimator instance
```
jhlegarreta, line 73: Same comment here.

src/eddymotion/validation.py
```diff
@@ -0,0 +1,112 @@
...
+
+def _is_fitted(estimator, attributes=None, all_or_any=all):
+    """Determine if an estimator is fitted.
+
+    Parameters
+    ----------
+    estimator : estimator instance
+        Estimator instance for which the check is performed.
+
+    attributes : str, list or tuple of str
```
jhlegarreta, line 37: Have not cared about the :obj:`str`, etc. for now. I'll fix them, once we figure out whether what is proposed is what was expected in the issue.

src/eddymotion/model/base.py
```diff
@@ -209,6 +215,10 @@ def predict(self, gradient, **kwargs):
 
         return retval
 
+    def __eddymotion_is_fitted__(self):
```
oesteban, line 218:
```suggestion
    def __model_is_fitted__(self):
```

src/eddymotion/validation.py
oesteban: I'd go with model. I agree the "fit" method of the estimator does not make much sense.

src/eddymotion/validation.py
```diff
@@ -0,0 +1,112 @@
...
+    -------
+    fitted : bool
+        Whether the estimator is fitted.
+    """
+    if attributes is not None:
+        if not isinstance(attributes, (list, tuple)):
+            attributes = [attributes]
+        return all_or_any([hasattr(estimator, attr) for attr in attributes])
+
+    return estimator.__eddymotion_is_fitted__()
```
oesteban, line 57:
```suggestion
    return estimator.__model_is_fitted__()
```

oesteban (approved):
I very much agree with the suggestions. I also agree that `fit()` was a very poor choice for the estimator object. Probably `optimize()` or `search()` would have been much better names. +1 to updating.

At this point, I'm not even sure the estimator should be an object at all.

src/eddymotion/validation.py
jhlegarreta: Changed in c440406.

src/eddymotion/validation.py
jhlegarreta: Changed in c440406.

src/eddymotion/validation.py
jhlegarreta: Done in c440406.

src/eddymotion/model/base.py
jhlegarreta: Changed in c440406.

src/eddymotion/validation.py
jhlegarreta: Changed in c440406.

jhlegarreta:
> eddymotion models (e.g. TrivialB0Model, etc.) do not inherit from BaseModel. Not sure if that is intended, but it is counter-intuitive or the rationale would need to be documented better.

Should they inherit from `BaseModel`?

> I very much agree with the suggestions. I also agree that fit() was a very poor choice for the estimator object. Probably optimize() or search() would have been much better names. +1 to updating.

Not sure if `optimize()` or `search()` are less confusing to me. If I read what its `fit` returns, `compute_transform` seems more descriptive and unambiguous wrt a fit/predict paradigm. To be addressed in a separate PR.

Concerning c440406, the `test_model` tests will not pass on CircleCI since the models do not inherit from `BaseModel`. If they are not supposed to inherit from it, I would need to relocate the `__model_is_fitted__` method or give the attributes to the `check_is_fitted` calls in `TrivialB0Model`, and `AverageDWModel`.

oesteban:
> Should they inherit from `BaseModel`?

I believe they do not inherit just for historical reasons (they were written the first, then we created `BaseModel` to provide a gateway for DIPY models and never did the last step of including them in the hierarchy). So, the answer is yes.

> Concerning [c440406](https://github.com/nipreps/eddymotion/commit/c440406fe996caa23a94c9b0b70c08ab172b70dd), the `test_model` tests will not pass on CircleCI since the models do not inherit from `BaseModel`. If they are not supposed to inherit from it, I would need to relocate the `__model_is_fitted__` method or give the attributes to the `check_is_fitted` calls in `TrivialB0Model`, and `AverageDWModel`.

Yup, I think making them part of the hierarchy makes all sense.

> Not sure if `optimize()` or `search()` are less confusing to me. If I read what its `fit` returns, `compute_transform` seems more descriptive and unambiguous wrt a fit/predict paradigm. To be addressed in a separate PR.

Yes, in a different PR :). `compute_transform` feels a bit wordy and even confusing (in the sense that we are actually computing a lot of transforms). What about `Estimator.estimate()`?

jhlegarreta:
> Yes, in a different PR :). compute_transform feels a bit wordy and even confusing (in the sense that we are actually computing a lot of transforms). What about Estimator.estimate()?

PR #173.

Making this a draft as addressing the inheritance issue is more complicated that what it seems.

oesteban (approved):

PR #171, opened 2024-04-16
ENH: Require Sphinx greater or equal 7.2

jhlegarreta:
Require Sphinx greater or equal 7.2 (released Aug 16, 2023).

Fixes:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed.
This behaviour is the source of the following dependency conflicts.
sphinx 4.5.0 requires docutils<0.18,>=0.14, but you have docutils 0.21.1 which is incompatible.
sphinx-rtd-theme 1.0.0 requires docutils<0.18, but you have docutils 0.21.1 which is incompatible.
```

raised for example in:
https://app.circleci.com/pipelines/github/nipreps/eddymotion/978/workflows/62e30eb5-72ed-4cf9-8b47-0172cad51228/jobs/951

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ doc = [
     "packaging",
     "pydot >= 1.2.3",
     "pydotplus",
-    "sphinx >= 4.5",
+    "sphinx >= 7.2",
     "sphinx-argparse",
     "sphinx_rtd_theme",
     "sphinxcontrib-apidoc ~= 0.3.0",


jhlegarreta:
**Untested**. None of my PRs has permissions to trigger CircleCI builds, and thus errors cannot be caught early enough:
https://support.circleci.com/hc/en-us/articles/360050273651-Troubleshooting-Builds-Unauthorized-due-to-contexts

esavary:
> **Untested**. None of my PRs has permissions to trigger CircleCI builds, and thus errors cannot be caught early enough: https://support.circleci.com/hc/en-us/articles/360050273651-Troubleshooting-Builds-Unauthorized-due-to-contexts

Hi, my PRs can trigger builds, but it seems I don't have write permissions on CircleCi to trigger yours. @oesteban or @effigies, could you trigger it for testing?

effigies:
@jhlegarreta I've invited you with write permissions on this repository. Hopefully Circle will figure it out. I manually triggered this one and got:

https://app.circleci.com/pipelines/github/nipreps/eddymotion/982/workflows/778a2cda-b4a4-418a-b59a-7cde7d3b40e5/jobs/962

```
INFO: pip is looking at multiple versions of eddymotion to determine which version is compatible with other requirements. This could take a while.
ERROR: Package 'eddymotion' requires a different Python: 3.9.12 not in '>=3.10'
```

jhlegarreta:
>  I've invited you with write permissions on this repository. Hopefully Circle will figure it out. I manually triggered this one and got:

Got that. Thanks Chris.

> I manually triggered this one and got:

OK. Not sure why it keeps on looking for 3.9: I first switched to a `cimg` docker image as the `nipreps` docker image had the 3.9 label on it, but then Oscar told me that the `nipreps` docker image was indeed shipping 3.11. So my best guess now is that the docker image we are using is not shipping or using Python >= 3.10.

effigies:
Right now we're using `nipreps/miniconda:py39_2205.0`, which bundles Python 3.9.

For reasons I don't understand, it looks like `py39_2403.0` does contain Python 3.11:

```console
❯ docker run --rm -it nipreps/miniconda:py39_2403.0 python --version
Python 3.11.8
```

jhlegarreta:
> For reasons I don't understand, it looks like py39_2403.0 does contain Python 3.11:

If multiple versions should be shipped, then we should make sure that it is picking Python 3.11, if there is a way to do that.

oesteban:
The reason is I copied parts of fMRIPrep's environment.yml and failed to
update the tag.

effigies:
> If multiple versions should be shipped, then we should make sure that it is picking Python 3.11, if there is a way to do that.

Update the tag in L8 of `.circleci/config.yml`.

jhlegarreta:
> Update the tag in L8 of .circleci/config.yml.

Which is the tag that should be used?

effigies:
py39_2403.0

jhlegarreta:
> py39_2403.0

:man_facepalming: yes, sorry, Chris: I now realize that your mentioned it. PR #177.

jhlegarreta:
Closing in favor of #177: looks like that is the right fix.

PR #173, opened 2024-04-16
STYLE: Rename estimator method to honor its purpose/use

jhlegarreta:
Rename estimator method to honor its purpose/use.


jhlegarreta:
Style issue addressed in PR #172.

oesteban (approved):

Issue #174, opened 2024-04-16
Should parallelization be handled upstream?


arokem:
This is a suggestion to simplify this: 

https://github.com/nipreps/eddymotion/blob/b1f70cc67417ac2fab43ccf5196b9b26397b2326/src/eddymotion/model/base.py#L160

We are in the midst of some work to parallelize model fitting in DIPY itself: https://github.com/dipy/dipy/pull/2593, which would benefit all downstream uses of these models, including here. I wonder whether we should aim to use these methods as implemented there (e.g., by passing through parallelization kwargs to DIPY's multi-voxel parallelization methods), instead of implementing that separately here.

oesteban:
Yes :)

arokem:
OK - well - let's keep this open if that's fine by you, so that we remember to refactor that code once the parallelization code is incorporated into DIPY (hopefully in the upcoming release).

arokem:
Good news! https://github.com/dipy/dipy/pull/2593 got merged 😄

arokem:
Actually, reopening, because we still need to remove the parallelization code from here.

PR #176, opened 2024-04-16
ENH: Make models inherit from base model

jhlegarreta:
Make models inherit from base model.

### DIFF
--- a/src/eddymotion/model/base.py
+++ b/src/eddymotion/model/base.py
@@ -415,7 +415,6 @@ class PETModel(BaseModel):
     @property
     def is_fitted(self):
         return self._coeff is not None
-        
     def fit(self, data, **kwargs):
         """Fit the model."""
         from scipy.interpolate import BSpline
--- a/src/eddymotion/model/base.py
+++ b/src/eddymotion/model/base.py
@@ -27,6 +27,7 @@ import warnings
 import numpy as np
 from dipy.core.gradients import gradient_table
 from joblib import Parallel, delayed
+
 from eddymotion.exceptions import ModelNotFittedError
 
 DEFAULT_MIN_S0 = 1e-5
@@ -56,9 +57,9 @@ def _exec_fit(model, data, chunk=None):
     return retval, chunk
 
 
-def _exec_predict_dwi(model, gradient, chunk=None, **kwargs):
+def _exec_predict(model, chunk=None, **kwargs):
     """Propagate model parameters and call predict."""
-    return np.squeeze(model.predict(gradient, S0=kwargs.pop("S0", None))), chunk
+    return np.squeeze(model.predict(**kwargs)), chunk
 
 
 class ModelFactory:
@@ -82,7 +83,7 @@ class ModelFactory:
 
         """
         if model.lower() in ("s0", "b0"):
-            return TrivialB0Model(S0=kwargs.pop("S0"))
+            return TrivialB0Model(S0=kwargs.pop("S0"), gtab=kwargs.pop("gtab"))
 
         if model.lower() in ("avg", "average", "mean"):
             return AverageDWModel(**kwargs)
@@ -111,25 +112,205 @@ class BaseModel:
         "_models",
         "_datashape",
         "_is_fitted",
+        "_modelargs",
     )
-    _modelargs = ()
 
     def __init__(self, mask=None, **kwargs):
         """Base initialization."""
 
-        self._model = None
+        # Keep model state
+        self._model = None  # "Main" model
+        self._models = None  # For parallel (chunked) execution
         self._is_fitted = False
 
         # Setup brain mask
         self._mask = mask
 
         self._datashape = None
-        self._models = None
         self._is_fitted = False
+
+        self._modelargs = ()
 
     @property
     def is_fitted(self):
         return self._is_fitted
+
+    def fit(self, data, **kwargs):
+        """Abstract member signature of fit()."""
+        raise NotImplementedError("Cannot call fit() on a BaseModel instance.")
+
+    def predict(self, *args, **kwargs):
+        """Abstract member signature of predict()."""
+        raise NotImplementedError("Cannot call predict() on a BaseModel instance.")
+
+
+class PETModel(BaseModel):
+    """A PET imaging realignment model based on B-Spline approximation."""
+
+    __slots__ = ("_t", "_x", "_xlim", "_order", "_coeff", "_n_ctrl")
+
+    def __init__(self, timepoints=None, xlim=None, n_ctrl=None, order=3, **kwargs):
+        """
+        Create the B-Spline interpolating matrix.
+
+        Parameters:
+        -----------
+        timepoints : :obj:`list`
+            The timing (in sec) of each PET volume.
+            E.g., ``[15.,   45.,   75.,  105.,  135.,  165.,  210.,  270.,  330.,
+            420.,  540.,  750., 1050., 1350., 1650., 1950., 2250., 2550.]``
+
+        n_ctrl : :obj:`int`
+            Number of B-Spline control points. If `None`, then one control point every
+            six timepoints will be used. The less control points, the smoother is the
+            model.
+
+        """
+        super.__init__(**kwargs)
+
+        if timepoints is None or xlim is None:
+            raise TypeError("timepoints must be provided in initialization")
+
+        self._order = order
+
+        self._x = np.array(timepoints, dtype="float32")
+        self._xlim = xlim
+
+        if self._x[0] < DEFAULT_TIMEFRAME_MIDPOINT_TOL:
+            raise ValueError("First frame midpoint should not be zero or negative")
+        if self._x[-1] > (self._xlim - DEFAULT_TIMEFRAME_MIDPOINT_TOL):
+            raise ValueError("Last frame midpoint should not be equal or greater than duration")
+
+        # Calculate index coordinates in the B-Spline grid
+        self._n_ctrl = n_ctrl or (len(timepoints) // 4) + 1
+
+        # B-Spline knots
+        self._t = np.arange(-3, float(self._n_ctrl) + 4, dtype="float32")
+
+        self._coeff = None
+
+    @property
+    def is_fitted(self):
+        return self._coeff is not None
+
+    def fit(self, data, **kwargs):
+        """Fit the model."""
+        from scipy.interpolate import BSpline
+        from scipy.sparse.linalg import cg
+
+        n_jobs = kwargs.pop("n_jobs", None) or 1
+
+        timepoints = kwargs.get("timepoints", None) or self._x
+        x = (np.array(timepoints, dtype="float32") / self._xlim) * self._n_ctrl
+
+        self._datashape = data.shape[:3]
+
+        # Convert data into V (voxels) x T (timepoints)
+        data = data.reshape((-1, data.shape[-1])) if self._mask is None else data[self._mask]
+
+        # A.shape = (T, K - 4); T= n. timepoints, K= n. knots (with padding)
+        A = BSpline.design_matrix(x, self._t, k=self._order)
+        AT = A.T
+        ATdotA = AT @ A
+
+        # One single CPU - linear execution (full model)
+        if n_jobs == 1:
+            self._coeff = np.array([cg(ATdotA, AT @ v)[0] for v in data])
+            return
+
+        # Parallelize process with joblib
+        with Parallel(n_jobs=n_jobs) as executor:
+            results = executor(delayed(cg)(ATdotA, AT @ v) for v in data)
+
+        self._coeff = np.array([r[0] for r in results])
+
+    def predict(self, index=None, **kwargs):
+        """Return the corrected volume using B-spline interpolation."""
+        from scipy.interpolate import BSpline
+
+        if index is None:
+            raise ValueError("A timepoint index to be simulated must be provided.")
+
+        if not self._is_fitted:
+            raise ModelNotFittedError(f"{type(self).__name__} must be fitted before predicting")
+
+        # Project sample timing into B-Spline coordinates
+        x = (index / self._xlim) * self._n_ctrl
+        A = BSpline.design_matrix(x, self._t, k=self._order)
+
+        # A is 1 (num. timepoints) x C (num. coeff)
+        # self._coeff is V (num. voxels) x K - 4
+        predicted = np.squeeze(A @ self._coeff.T)
+
+        if self._mask is None:
+            return predicted.reshape(self._datashape)
+
+        retval = np.zeros(self._datashape, dtype="float32")
+        retval[self._mask] = predicted
+        return retval
+
+
+class BaseDWIModel(BaseModel):
+    """Interface and default methods for DWI models."""
+
+    __slots__ = (
+        "_gtab",
+        "_S0",
+        "_b_max",
+        "_model_class",  # Defining a model class, DIPY models are instantiated automagically
+        "_modelargs",
+    )
+
+    def __init__(self, gtab, S0=None, b_max=None, **kwargs):
+        """Initialization.
+
+        Parameters
+        ----------
+        gtab : :obj:`numpy.ndarray`
+            An :math:`N \times 4` table, where rows (*N*) are diffusion gradients and
+            columns are b-vector components and corresponding b-value, respectively.
+        S0 : :obj:`numpy.ndarray`
+            :math:`S_{0}` signal.
+        b_max : :obj:`int`
+            Maximum value to cap b-values.
+
+        """
+
+        super().__init__(**kwargs)
+
+        # Setup B0 map
+        self._S0 = None
+        if S0 is not None:
+            self._S0 = np.clip(
+                S0.astype("float32") / S0.max(),
+                a_min=DEFAULT_MIN_S0,
+                a_max=DEFAULT_MAX_S0,
+            )
+
+        # Cap b-values, if requested
+        self._gtab = gtab
+        self._b_max = None
+        if b_max and b_max > DEFAULT_MAX_BVALUE:
+            # Saturate b-values at b_max, since signal stops dropping
+            self._gtab[-1, self._gtab[-1] > b_max] = b_max
+            # A possibly good alternative is completely remove very high b-values
+            # bval_mask = gtab[-1] < b_max
+            # data = data[..., bval_mask]
+            # gtab = gtab[:, bval_mask]
+            self._b_max = b_max
+
+        kwargs = {k: v for k, v in kwargs.items() if k in self._modelargs}
+
+        # DIPY models (or one with a fully-compliant interface)
+        model_str = getattr(self, "_model_class", None)
+        if model_str:
+            from importlib import import_module
+
+            module_name, class_name = model_str.rsplit(".", 1)
+            self._model = getattr(
+                import_module(module_name),
+                class_name,
+            )(_rasb2dipy(gtab), **kwargs)
 
     def fit(self, data, n_jobs=None, **kwargs):
         """Fit the model chunk-by-chunk asynchronously"""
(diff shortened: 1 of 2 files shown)


jhlegarreta:
@effigies @oesteban Following this comment https://github.com/nipreps/eddymotion/pull/166#issuecomment-2058387340, I had a look at inheriting the models from `ModelBase`. Things were unclear, and I ended up adding my questions as comments in the code. Not good practice for reviewing, but would be grateful if you had a look and commented.

The main point is that the `BaseClass`' `fit` and `predict` methods contain a lot of code that is not being call at all. The initialization method is not used either. The latter may be easy to fix with a call to the superclass init from the child classes, but it is unclear how the code in the `fit` and `predict` methods of `BaseClass` should be reused.

This is also related to issue #174.

So comments would be appreciated.

src/eddymotion/model/base.py
```diff
@@ -166,12 +177,15 @@ def fit(self, data, n_jobs=None, **kwargs):
 
         self._model = None  # Preempt further actions on the model
 
-    def predict(self, gradient, **kwargs):
+    def predict(self, **kwargs):
```
oesteban, line 180: Predict requires you tell the model which index you want to infer.

When models were DWI-only, this had the `gradient`, so you pass the direction. But, if it needs to generalize, I like your idea of having the gradient table store somewhere and just pass the index you want to simulate:

```suggestion
    def predict(self, index, **kwargs):
```

src/eddymotion/model/base.py
```diff
@@ -87,13 +87,16 @@ class BaseModel:
         "_model",
         "_mask",
         "_S0",
+        "_gtab",
```
oesteban, line 90: This is very specific of DWI - happy to move forward with it, but we may end up increasing the inheritance chain to have a more general abstract interface.

(this also applies to `_S0`, which is just one line above and was here before the PR, and same for `_b_max` below)

src/eddymotion/model/base.py
```diff
@@ -287,12 +304,12 @@ def fit(self, data, **kwargs):
...
         self._data = avg_func(shells, axis=-1)
 
-    def predict(self, gradient, **kwargs):
+    def predict(self, **kwargs):
         """Return the average map."""
         return self._data
 
 
-class PETModel:
+class PETModel:  ## Should this inherit from BaseModel ??
```
oesteban, line 312: Okay, you were already advancing the issue I mention above. We do want a `BaseModel` for all possible models, so maybe what's missing is a `BaseDWIModel`

src/eddymotion/model/base.py
```diff
@@ -368,8 +385,10 @@ def fit(self, data, *args, **kwargs):
 
         self._coeff = np.array([r[0] for r in results])
 
-    def predict(self, timepoint, **kwargs):
-        """Return the *b=0* map."""
+    def predict(
+        self, timepoint, **kwargs
```
oesteban, line 389: Following the pattern proposed above:
```suggestion
        self, index, **kwargs
```

oesteban (commented):
I think this will make the code much more readable -- left some comments.

src/eddymotion/model/base.py
jhlegarreta: OK. See comments in commit df51602. Resolving.

src/eddymotion/model/base.py
jhlegarreta: Done. See comments in commit df51602. Resolving.

src/eddymotion/model/base.py
jhlegarreta: OK. See comments in commit df51602. Resolving.

src/eddymotion/model/base.py
jhlegarreta: Done in commit df51602. Resolving.

src/eddymotion/model/base.py
```diff
@@ -28,13 +28,21 @@
 from dipy.core.gradients import gradient_table
 from joblib import Parallel, delayed
 
+a_min_S0 = 1e-5
+a_max_S0 = 1.0
+bval_upper_cap = 1000
+bval_lower_th = 50
+bval_upper_th = 10000
+percentil_percentage = 75
+timeframe_midpoint_tol = 1e-2  # in seconds
```
jhlegarreta, line 37: I declared this variables at the global scope, but should they be made method parameters ?

src/eddymotion/model/base.py
```diff
@@ -166,12 +141,73 @@ def fit(self, data, n_jobs=None, **kwargs):
 
         self._model = None  # Preempt further actions on the model
```
jhlegarreta, line 142: Not sure if I understand this.

src/eddymotion/model/base.py
```diff
@@ -28,13 +28,21 @@
...
+timeframe_midpoint_tol = 1e-2  # in seconds
+
 
 def _exec_fit(model, data, chunk=None):
     retval = model.fit(data)
     return retval, chunk
 
 
-def _exec_predict(model, gradient, chunk=None, **kwargs):
+def _exec_predict_dwi(model, gradient, chunk=None, **kwargs):
```
jhlegarreta, line 45: I renamed this method to contain the `dwi` label, as it requires a gradient and optionally uses a `S0` argument. If `gradient` in reality should be an `index`, it may be renamed back. Also, as things are right now, I do not see the need to pass an `S0` either.

src/eddymotion/model/base.py
```diff
@@ -210,27 +246,25 @@ def predict(self, gradient, **kwargs):
...
+    def fit(self, data, **kwargs):
         """Do nothing."""
 
-    def predict(self, gradient, **kwargs):
+    def predict(self, *_, **kwargs):
         """Return the *b=0* map."""
         return self._S0
 
 
-class AverageDWModel:
```
jhlegarreta, line 233: Why was the `I` from `DWI` droped for `AverageDWModel` (vs `AverageDWIModel` (or vs `dMRI`, `DMRI`).

This dwi vs dmri naming discrepancy (if it is such) is also present in other place (and other repositories, e.g. `nipreps`). At some point a decision will need to be made to make things consistent (or the policy be explained if there is one).

src/eddymotion/model/base.py
```diff
@@ -166,12 +141,73 @@ def fit(self, data, n_jobs=None, **kwargs):
...
+        self._model = getattr(import_module(module_name), class_name)(_rasb2dipy(gtab), **kwargs)
+
+    def predict(self, index, **kwargs):
         """Predict asynchronously chunk-by-chunk the diffusion signal."""
         if self._b_max is not None:
-            gradient[-1] = min(gradient[-1], self._b_max)
+            index[-1] = min(index[-1], self._b_max)
 
-        gradient = _rasb2dipy(gradient)
+        self._gtab = _rasb2dipy(self._gtab)
```
jhlegarreta, line 210: Not sure about this: `gradient` contains the whole gtab according to `_ras2dipy`.

See also https://github.com/nipreps/eddymotion/pull/176/files#diff-a875f501910044a7d95658fb83740e2c5c6c1693e7e6808703d282441db82be8R203

src/eddymotion/model/base.py
```diff
@@ -166,12 +141,73 @@ def fit(self, data, n_jobs=None, **kwargs):
...
+        from importlib import import_module
+
+        module_name, class_name = model_str.rsplit(".", 1)
+        self._model = getattr(import_module(module_name), class_name)(_rasb2dipy(gtab), **kwargs)
+
+    def predict(self, index, **kwargs):
         """Predict asynchronously chunk-by-chunk the diffusion signal."""
         if self._b_max is not None:
-            gradient[-1] = min(gradient[-1], self._b_max)
+            index[-1] = min(index[-1], self._b_max)
```
jhlegarreta, line 208: I think this is not OK. If the gradients are capped, not sure how the indices get affected/how they should be checked.

src/eddymotion/model/base.py
```diff
@@ -241,43 +275,41 @@ def __init__(self, **kwargs):
...
     def fit(self, data, **kwargs):
         """Calculate the average."""
-        gtab = kwargs.pop("gtab", None)
         # Select the interval of b-values for which DWIs will be averaged
         b_mask = (
-            ((gtab[3] >= self._th_low) & (gtab[3] <= self._th_high))
-            if gtab is not None
+            ((self._gtab[3] >= self._th_low) & (self._gtab[3] <= self._th_high))
+            if self._gtab is not None
             else np.ones((data.shape[-1],), dtype=bool)
```
jhlegarreta, line 305: Unless we model DSI acquisitions `self._gtab` will not be none, so this `if/else` block is not necessary to me.

src/eddymotion/model/base.py
```diff
@@ -86,51 +94,18 @@ class BaseModel:
     __slots__ = (
         "_model",
         "_mask",
-        "_S0",
-        "_b_max",
         "_models",
```
jhlegarreta, line 97: What is the difference between `model` and `models`? Unclear to me.

jhlegarreta:
Gave this another go. Adjusted some docstrings.

More questions/comments (inline and below, long to digest, sorry):

- I am not sure whether the parent slots are inherited by the child classes
- I am not sure, and how the `AverageDWModel` `__init__` method `kwargs` documentation is understood, as (i) it only contains kwargs (we may need to tell something like `kwargs can contain...` - not sure how this is done properly in Sphinx); (ii) the parameters are defaulted to some values if not found. So maybe they can just be list as regular keyword arguments.
- Not sure if I follow the timepoint/index rationale for the PET model. Do we have multiple PET volumes for a single session? If we do not, then I am not sure to follow: this would be like requiring DWI volumes from different sessions/timepoints to be aligned, which is different from what we try to do for the DWI case (correcting data from the same session).
- Following [this comment about DSI](https://github.com/nipreps/mriqc/pull/1292#issuecomment-2084966496), the requirement of a `gtab` in `BaseDWIModel` may need to be changed (or create a Cartesian vs Spherical acq base classes). IMHO, this should be addressed in a separate PR, otherwise the PR will get stalled.
- The following:
https://github.com/nipreps/eddymotion/pull/176/files#diff-a875f501910044a7d95658fb83740e2c5c6c1693e7e6808703d282441db82be8R117

  Looks like it does not apply to PET (see https://github.com/nipreps/eddymotion/pull/176/files#diff-a875f501910044a7d95658fb83740e2c5c6c1693e7e6808703d282441db82be8R382). So should it be moved to the DWI base class?

jhlegarreta:
Sorry to ping you again his morning @oesteban.

src/eddymotion/model/base.py
jhlegarreta: According to an in-person conversation, we may need `models` to keep the model for each chunk of data that is processed. Keeping for now. Resolving.

src/eddymotion/model/base.py
jhlegarreta: Can be addressed at a later time.  For now, the relevant thing is to have a proper base model to inherit from. Resolving.

src/eddymotion/model/base.py
```diff
@@ -86,51 +94,18 @@ class BaseModel:
...
-            # bval_mask = gtab[-1] < b_max
-            # data = data[..., bval_mask]
-            # gtab = gtab[:, bval_mask]
-            self._b_max = b_max
-
-        kwargs = {k: v for k, v in kwargs.items() if k in self._modelargs}
-
-        model_str = getattr(self, "_model_class", None)
-        if not model_str:
-            raise TypeError("No model defined")
```
jhlegarreta, line 128: @oesteban Some tests are now failing because the `_model_class` is `None` in this base class, and I am not setting any particular value in the derived classes. What is this property supposed to contain?

e.g.
https://app.circleci.com/pipelines/github/nipreps/eddymotion/1070/workflows/c9747d35-0cb1-49d9-963f-207d20887ce8/jobs/1038

src/eddymotion/model/base.py
jhlegarreta: I think there is a naming confusion here: the models expect a RAS+b gradient object (which is different from the `dipy` `gtab` object) into a `gtab` parameter. Am I correct @oesteban ?

src/eddymotion/model/base.py
jhlegarreta: I am confused about the use of this. I now see that `_model_class` and `_modelargs` are properties of the wrappers `DTIModel` and `DKIModel`. This adds to this docstring https://github.com/nipreps/eddymotion/pull/176/files#diff-a875f501910044a7d95658fb83740e2c5c6c1693e7e6808703d282441db82be8L79

If I say here
```
kwargs = {k: v for k, v in kwargs.items() if k in self._modelargs}
from importlib import import_module
model_str = "eddymotion.model.AverageDWModel"
module_name, class_name = model_str.rsplit(".", 1)
my_model = getattr(import_module(module_name), class_name)(**kwargs)
```

and leaving aside that `AverageDWModel` requires a `gtab` (as it inherits from `BaseDWIModel`; setting it to `None` in its `init` would make it), the above statement produces a recursive call, since instantiating `AverageDWModel` calls the superclass `init` method.

So I am not following what was intended with this block.

Also, I am not sure what we want to do here with the DTI and DKI wrappers either.

@oesteban Can you please clarify this?

jhlegarreta: > and leaving aside that AverageDWModel requires a gtab (as it inherits from BaseDWIModel;

The other two tests fail because of this reason.

src/eddymotion/model/base.py
oesteban: These are constants, so I would name them in all caps and set some clear prefix for all of them:

```suggestion
DEFAULT_MIN_S0 = 1e-5
DEFAULT_MAX_S0 = 1.0
DEFAULT_MAX_BVALUE = 1000
DEFAULT_LOWB_THRESHOLD = 50
DEFAULT_HIGHB_THRESHOLD = 10000
DEFAULT_CLIP_PERCENTILE = 75
DEFAULT_TIMEFRAME_MIDPOINT_TOL = 1e-2  # in seconds
```

src/eddymotion/model/base.py
oesteban: I'll have a look ASAP - sorry for my slow turnaround

oesteban (commented):
I have to checkout this code to make a better review of it. A nit pick for the time being.

src/eddymotion/model/base.py
jhlegarreta: Done in 324d2ee.

oesteban:
Let's get #166 over the final line and then I move onto this.

src/eddymotion/model/base.py
```diff
@@ -27,16 +27,36 @@
 import numpy as np
 from dipy.core.gradients import gradient_table
 from joblib import Parallel, delayed
-
 from eddymotion.exceptions import ModelNotFittedError
 
+#: Minimum value when considering the :math:`S_{0}` DWI signal.
+DEFAULT_MIN_S0 = 1e-5
```
oesteban, line 33: Python docstrings can (should) be used to document constants:
```suggestion
DEFAULT_MIN_S0 = 1e-5
"""Minimum value when considering the :math:`S_{0}` DWI signal."""
```

src/eddymotion/model/base.py
```diff
@@ -27,16 +27,36 @@
...
+
+#: Maximum value when considering the :math:`S_{0}` DWI signal.
+DEFAULT_MAX_S0 = 1.0
```
oesteban, line 36:
```suggestion
DEFAULT_MAX_S0 = 1.0
"""Maximum value when considering the :math:`S_{0}` DWI signal."""
```

src/eddymotion/model/base.py
```diff
@@ -27,16 +27,36 @@
...
+
+#: Maximum allowed value for the b-value.
+DEFAULT_MAX_BVALUE = 1000
```
oesteban, line 39:
```suggestion
DEFAULT_MAX_BVALUE = 1000
"""Maximum allowed value for the b-value."""
```

src/eddymotion/model/base.py
```diff
@@ -27,16 +27,36 @@
...
+
+#: b-value lower bound when considering DWI data.
+DEFAULT_LOWB_THRESHOLD = 50
```
oesteban, line 42:
```suggestion
DEFAULT_LOWB_THRESHOLD = 50
"""The lower bound for the b-value so that the orientation is considered a DW volume."""
```

src/eddymotion/model/base.py
```diff
@@ -27,16 +27,36 @@
...
+
+#: b-value upper bound when considering DWI data.
+DEFAULT_HIGHB_THRESHOLD = 10000
```
oesteban, line 45:
```suggestion
DEFAULT_HIGHB_THRESHOLD = 10000
"""A b-value cap for DWI data."""
```

src/eddymotion/model/base.py
```diff
@@ -27,16 +27,36 @@
...
+
+#: Percentile clipping value.
+DEFAULT_CLIP_PERCENTILE = 75
```
oesteban, line 48:
```suggestion
DEFAULT_CLIP_PERCENTILE = 75
"""Upper percentile threshold for intensity clipping."""
```

src/eddymotion/model/base.py
```diff
@@ -27,16 +27,36 @@
...
+
+#: Time frame tolerance in seconds.
+DEFAULT_TIMEFRAME_MIDPOINT_TOL = 1e-2
```
oesteban, line 51: PET, I assume?
```suggestion
DEFAULT_TIMEFRAME_MIDPOINT_TOL = 1e-2
"""Time frame tolerance in seconds."""
```

oesteban (commented):
I like where this is going. I'm going to add the docstring of constants and then work locally on this PR.

src/eddymotion/model/base.py
oesteban: Sorry for the delayed response. I can now answer this question. I'm sorry for this particular case—you are prey to an undocumented feature.

The idea was that models can be fit in two ways:
- Pure leave-one-out fashion: at every iteration of the Estimator, a fully-fledged model is fit without the particular index/orientation. This is typically very slow.
- Single model: the model is fit on all the data, and each iteration produces the left-out index. These are enable by adding the prefix `Full` to the model name.

This is implemented in the Estimator, under the understanding that the model is the same, what changes is how you use it.
https://github.com/nipreps/eddymotion/blob/ce8de17b63ecce46b132d6a14d3c3ab34de3cca4/src/eddymotion/estimator.py#L117-L135

ATM I cannot comment on why this has some effect on the model itself so a `_model_class` is necessary, I can't recall the reason. I bet it is just to inform the estimator that fit should not be called every time (which probably should be handled here!)

That said, let's take the average model for example. When instantiated as `FullAverage`, then it is fit only once before entering the iterator loop of the estimator. If not, at every iteration an average without the particular direction will be calculated in the fit call.

src/eddymotion/model/base.py
```diff
@@ -366,14 +410,13 @@ def __init__(self, timepoints=None, xlim=None, n_ctrl=None, mask=None, order=3,
...
 
-        self._shape = None
         self._coeff = None
 
     @property
     def is_fitted(self):
         return self._coeff is not None
-
-    def fit(self, data, *args, **kwargs):
+
```
oesteban, line 418:
```suggestion

```

oesteban (commented):
I responded to the two major questions in this PR. Happy to chat about the `_model_class` as it seems the feature may not be implemented in an intuitive way and it is definitely not sufficiently documented.

jhlegarreta:
@oesteban Have gone through the comments. Will wait after this https://github.com/nipreps/eddymotion/pull/176#discussion_r1631915776.

The main difficulty to make this work now lies in https://github.com/nipreps/eddymotion/pull/176/files#r1616403642. Although you answered to the thread, not sure if the question was addressed: the point is that I do not see why `DTIModel` and `DKIModel` exist in here; if they are meant to be removed, the issue related to `_model_class` would go away, and the inheritance would be easier I think, as it is the e.g. `TrivialB0Model` and `AverageDWModel` DWI models the ones that we are interested in subclassing from `BaseDWIModel`<-`BaseModel`.

src/eddymotion/model/base.py
oesteban: This is capping the b value (only the last item of `gradient`. For some models, very high b-values 'saturate' and it's better to model as if they were lower. This only kicks in after setting b_max so you need to be explicit about it.

src/eddymotion/model/base.py
oesteban: Here we do want to use the input gtab, as opposed to the global gtab, which potentially contains the left-out gradient.

src/eddymotion/model/base.py
oesteban: Okay, I'll leave my above comment because it explains something useful --- but it is totally unrelated to @jhlegarreta's question. Apologies for the confusion.

After working on the PR and re-reading the code, I understand that `_model_class` and `_modelargs` enable using DIPY models without much overhead (see DKI and DTI at the end).

jhlegarreta:
@oesteban Had to adjust the test:
https://github.com/nipreps/eddymotion/pull/176/files#diff-b7713b8731a4d1f895063e58f31f8ab54f7bb6ca0e2b1712d85683874d94a6e6R44-R57

due to this now being applied to the `TrivialB0Model`:
https://github.com/nipreps/eddymotion/pull/176/files#diff-a875f501910044a7d95658fb83740e2c5c6c1693e7e6808703d282441db82be8R284-R287

src/eddymotion/model/base.py
```diff
@@ -177,16 +344,21 @@ def fit(self, data, n_jobs=None, **kwargs):
...
 
-        if self._b_max is not None:
-            gradient[-1] = min(gradient[-1], self._b_max)
+        gradient = np.array(gradient)  # Tuples are unmutable
+
+        # Cap the b-value if b_max is defined
+        gradient[-1] = min(gradient[-1], self._b_max or gradient[-1])
 
-        gradient = _rasb2dipy(gradient)
+        self._gtab = _rasb2dipy(gradient)
```
oesteban, line 361: We should not override the gradient table in predict:

```suggestion
        gradient = _rasb2dipy(gradient)
```

src/eddymotion/model/base.py
```diff
@@ -199,7 +371,7 @@ def predict(self, gradient, **kwargs):
         n_models = len(self._models) if self._model is None and self._models else 1
 
         if n_models == 1:
-            predicted, _ = _exec_predict(self._model, gradient, S0=S0, **kwargs)
+            predicted, _ = _exec_predict(self._model, **(kwargs | {"gtab": self._gtab, "S0": S0}))
```
oesteban, line 374:
```suggestion
            predicted, _ = _exec_predict(self._model, **(kwargs | {"gtab": gradient, "S0": S0}))
```

src/eddymotion/model/base.py
```diff
@@ -208,7 +380,11 @@ def predict(self, gradient, **kwargs):
             # Parallelize process with joblib
             with Parallel(n_jobs=n_models) as executor:
                 results = executor(
-                    delayed(_exec_predict)(model, gradient, S0=S0[i], chunk=i, **kwargs)
+                    delayed(_exec_predict)(
+                        model,
+                        chunk=i,
+                        **(kwargs | {"gtab": self._gtab, "S0": S0[i]}),
```
oesteban, line 386:
```suggestion
                        **(kwargs | {"gtab": gradient, "S0": S0[i]}),
```

oesteban (approved):

Issue #152, opened 2024-04-03
Release 0.1.15


oesteban:
We could've released it long ago, as a fair amount of features have been included. Let's exercise the release process with the current status.

esavary:
I've drafted the release [here](https://github.com/nipreps/eddymotion/releases/tag/untagged-6eb1689f8bdc2850c95a). Should we perhaps merge #171 and #172 before releasing in order to include the fixes for the ruff and Sphinx issues? Also, a naive question: Is there a mechanism to automatically generate the binaries for the release, or is it okay if I drag and drop some generated locally?

arokem:
I think that adding something like [this](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) would be beneficial.

esavary:
> I think that adding a GitHub action for generating the package and uploading it to pypi (something like [this](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)) would be beneficial.
> 
> EDIT: edited for clarity.

It seems like a good idea. I've created a new issue about it and will start working a PR to implement this.

esavary:
> We normally do this with circle, paired with uploading a docker image -- in this case that doesn't seem necessary.
> 
> The other reason for doing it from circle is that we do run more tests there (we pay credits for it) and therefore, a GHA may push a new release with tests failing.

Ok thanks, I've close the issue related to this #178 . Can you provide more details on how I can use this to add the files for [the release](https://github.com/nipreps/eddymotion/releases/tag/untagged-6eb1689f8bdc2850c95a)?

esavary:
@effigies considering that #182 and #179 for the automated releases have been merged, do you think we can proceed with cutting the release?

effigies:
Yes, please don't wait on me. LMK if you need help.

esavary:
> Yes, please don't wait on me. LMK if you need help.

I have a quick question: should I update eddymotion/CHANGES.rst before pushing the tag? The file seems to have been automatically generated during the initial release, but I'm not entirely sure how it works.

effigies:
For a given tag (`$TAG`):

```console
bash .maint/update_changes.sh $TAG
# review CHANGES.rst, make any updates for clarity, add release notes
# I generally group the entries by FIX, ENH, RF, DOC, STY, MNT, CI
git add CHANGES.rst
git commit -m "REL: $TAG"
[...]

PR #179, opened 2024-04-24
ENH: Automated Package Building and Deployment

esavary:
This PR introduces new jobs in the build_test_deploy CircleCI workflow for building and deploying the package to PyPI.
related to #152


esavary:
@effigies  @oesteban , I'm wondering if there's a clever way to test the code I added in the CircleCi workflow without needing to push a tag. (To prevent misunderstanding, @oesteban , you mentioned yesterday using TestPyPI to test the package, but what I'm asking here is a bit different—I'm looking for a way to specifically test the new CircleCI routine itself, not the package.)  Any suggestions on this would be helpful!

effigies:
For new workflows, I would strongly recommend adopting https://github.com/hynek/structlog/blob/main/.github/workflows/pypi-package.yml more or less wholesale. Its first action produces easily inspectable reports in the actions tab: https://github.com/hynek/structlog/actions/runs/8584514099

image

Since these actions are maintained, they are less likely to bitrot than custom scripts in CircleCI.

effigies:
We can create a deployment environment that requires a maintainer to approve the release before pushing.

effigies:
Here is an example of using an environment:

https://github.com/nipy/nibabel/blob/a2e5dee05cf374c22670ff9fd0d385ce366eb495/.github/workflows/test.yml#L199-L212

I created one here called "Release"

esavary:
> Here is an example of using an environment:
> 
> https://github.com/nipy/nibabel/blob/a2e5dee05cf374c22670ff9fd0d385ce366eb495/.github/workflows/test.yml#L199-L212
> 
> I created one here called "Release"

Thanks a lot for the example, I'll move forward in this direction and convert my script to use GitHub Actions.

effigies:
@oesteban Can you make me an admin on the eddymotion PyPI package?

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,72 @@
...
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist/
+      - uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: https://test.pypi.org/legacy/
+          user: __token__
+          password: ${{ secrets.TESTPYPI_API_TOKEN }}
```
esavary, line 55: @effigies @oesteban Do we also have an authentication token for TestPyPI? (For now, I've used this generic variable name.)

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,72 @@
...
+    - run: pip install --upgrade build twine
+    - name: Build sdist and wheel
+      run: python -m build
+    - name: Check package
+      run: twine check dist/*
+    - name: Upload sdist and wheel artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: dist
+        path: dist/
```
effigies, line 39:
```suggestion
    name: Build & verify package
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
      - uses: hynek/build-and-inspect-python-package@v2
```

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,72 @@
...
+  auto-release-test-pypi:
+    runs-on: ubuntu-latest
+    environment: "Test-release"
+    needs: [build-package]
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist/
```
effigies, line 50:
```suggestion
      - name: Download packages built by build-and-inspect-python-package
        uses: actions/download-artifact@v4
        with:
          name: Packages
          path: dist
```

.github/workflows/pypi-package.yml
effigies, line 55: Let's just drop these. I've set up [trusted publishing](https://docs.pypi.org/trusted-publishers/) on test.pypi.org.
```suggestion
```

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,72 @@
...
+      run: twine check dist/*
+    - name: Upload sdist and wheel artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: dist
+        path: dist/
+
+  auto-release-test-pypi:
+    runs-on: ubuntu-latest
+    environment: "Test-release"
```
effigies, line 43: We don't need an environment for test pushes. It either does nothing or spams maintainers for approval on every push to `main`.

```suggestion
```

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,72 @@
...
+          repository-url: https://test.pypi.org/legacy/
+          user: __token__
+          password: ${{ secrets.TESTPYPI_API_TOKEN }}
+
+  auto-release-pypi:
+    runs-on: ubuntu-latest
+    environment: "Release"
+    needs: [build]
+    permissions:
+      id-token: write
```
effigies, line 62: This is already set globally.

```suggestion
```

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,72 @@
...
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist/
```
effigies, line 68:
```suggestion
      - name: Download packages built by build-and-inspect-python-package
        uses: actions/download-artifact@v4
        with:
          name: Packages
          path: dist
```

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,72 @@
...
+      - uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN }}
```
effigies, line 72: If given admin rights (cc @oesteban) on the PyPI package, I will set up trusted publishing for this, as well.

```suggestion
```

oesteban: Done, I believe

.github/workflows/pypi-package.yml
esavary: Thanks, indeed, it makes sense. I wasn't sure if we should also restrict the test releases.

esavary:
Is there a clever way to trigger the workflow for testing without pushing changes to the main branch?

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: hynek/build-and-inspect-python-package@v2
+
+  auto-release-test-pypi:
+    runs-on: ubuntu-latest
+    needs: [build-package]
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
effigies, line 30: > Is there a clever way to trigger the workflow for testing without pushing changes to the main branch?

Comment this out for now:

```suggestion
    # if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,52 @@
+---
+name: Build and upload package
+
+on:
+  push:
+    branches: [main]
+    tags: ["*"]
```
effigies, line 7: Looks like we also need:

```suggestion
  push:
    branches: [main]
    tags: ["*"]
  pull_request:
    branches: [ main ]
```

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,54 @@
...
+          name: Packages
+          path: dist
+      - uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: https://test.pypi.org/legacy/
+
+  auto-release-pypi:
+    runs-on: ubuntu-latest
+    environment: "Release"
+    needs: [build]
```
effigies, line 46:
```suggestion
    needs: [build-package]
```

effigies:
Okay, looks like we can't do this from a fork, but it's mostly working.

We will need to adjust our version string, though: test.pypi.org will not accept local identifiers like the `+gcc9a910` in `24.0.0.dev277+gcc9a910`

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: hynek/build-and-inspect-python-package@v2
+
+  auto-release-test-pypi:
+    runs-on: ubuntu-latest
+    needs: [build-package]
+    # if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
effigies, line 32:
```suggestion
    # if: github.event_name == 'push' && github.ref == 'refs/heads/main'
    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'nipreps/eddymotion' }}

```

effigies:
I'm going to merge this. We can fiddle with versions and then resetting the run conditions in a PR from the main repo.

effigies (approved):

PR #182, opened 2024-04-26
MNT: Drop local version to play nicely with test.pypi.org

effigies:
https://github.com/nipreps/eddymotion/actions/runs/8850953249/job/24306415105 failed because of the local version.


effigies:
Okay, the test passes: https://github.com/nipreps/eddymotion/actions/runs/8851070515?pr=182

Removing the PR builds.

jhlegarreta (approved):

PR #187, opened 2024-04-30
ENH: Call the gradient plot method from `nireports`

jhlegarreta:
Call the gradient plot method from its new location in `nireports`.


jhlegarreta:
Depends on https://github.com/nipreps/nireports/pull/96 and a new release of `nireports`.

oesteban:
@arokem @teresamg this is an example of the easy route to move the carpet plot into nireports (see nireports counterpart at nipreps/nireports#96).

Before @jhlegarreta I could have forwarded the pertinent commits over there, but with the new history, it makes more sense that teresa does it herself (so the credit goes to her).

effigies (approved):

PR #188, opened 2024-05-05
ENH: Implement Gaussian Process

jhlegarreta:
Implement Gaussian Process.


yibeichan:
I made a draft PR on @jhlegarreta 's fork here https://github.com/jhlegarreta/eddymotion/pull/1
I updated `calculate_angle_matrix` in [src/eddymotion/model/utils.py](https://github.com/yibeichan/eddymotion/blob/672c6b9556169360a8f1dfbaf44f925aa963935c/src/eddymotion/model/utils.py#L142)
but I realize that my PR has some conflicts that need to be merged due to the latest updates in `nippers/eddymotion`. It's probably too much work to merge such huge conflicts just for one function.
I explained to Jon why I updated the function so Jon can decide whether you want to adopt the changes (by copying it from the link above)

Regarding `compute_multi_shell_covariance_matrix`, I tried to work on it but had some problems
1. the input parameters in the current version of the function is not complete, for example `nvals` and `bvecs` needs to be defined
2. also, the current `calculate_angle_matrix` only works for single-shell data, so we need to update this function first and then call it in `compute_multi_shell_covariance_matrix`
3. also, make sure we are consistent on whether to use `k` as input in those covariance matrices or initiate `k` within the function.
4. Some parameter names are not consistent. For example, `theta` should be `angle_mat`, but we use different names across functions.

Thank you, @jhlegarreta, for explaining everything to me and getting me started on this work (which is different from what I've done before). I'm happy to contribute more in the future if needed.

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+    \beta: four for linear; ten for quadratic, twenty for cubic.
+    h: assumed to be in the same space as the first b = 0 image supplied to
+    eddy, which will be automatically fulfilled if it was estimated by topup
+    and that same b = 0 image was the first of those supplied to topup. Hence,
+    it can be said to help define the reference/undistorted space as the first
+    b = 0 image after distortion correction by h.
+
+    See Appendix A for further details.
+
+    Add the outlier detection part in [Andersson16b]?
```
oesteban, line 463: We don't for the moment.

```suggestion
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
 
+class GaussianProcessModel:
+    """A Gaussian Process model based on [Andersson16a]_ (fig 1).
+    DWIs need to be transformed to a single ref space (fig 2 [Andersson16b]_ ?)
+
+    Definitions:
+    s: reference/undistorted space: used to denote the space or any image in
+    that space
+    f: observed/distorted image: used to denote any image in acquisition space
+    a: acquisition parameters: PE‐direction and bandwidth in PE‐direction
```
oesteban, line 423: I think these only influence the susceptibility distortion, remove for now?

```suggestion
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+    r: rigid body (subject movement) parameters
+    \beta: Eddy current parameters
+    e(\beta): Eddy current‐induced off resonance field (Hz)
+    h: Susceptibility induced off‐resonance field (Hz)
```
oesteban, line 427: Not considered for now
```suggestion
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+
+    (fig 1) and algorithm:
+    1. Input: N DWI volumes f_{i} with acq parameters a_{i}; susceptibility
+    field h
```
oesteban, line 431:
```suggestion
    1. Input: N DWI volumes f_{i}, beta_{i}, r_i{i}, and hyperparameters.
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+    2. Initialize: set all beta_{i} and r_i{i} = 0
```
oesteban, line 432: This is the estimator's job, not the model AFAICT
```suggestion
    2. Initialize: set all beta_{i} and r_i{i} (passed in by the estimator)
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+    3. Compute for M iterations
```
oesteban, line 433: This is the estimator's job
```suggestion
    3. Compute for M iterations
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+       - Compute \\hat{s}_{i} (f_{i}, h, \beta_{i}, r_{i}, a_{i}) eqs 2 and 4
+       - Load \\hat{s}_{i} (f_{i}, h, \beta_{i}, r_{i}, a_{i}) as training
+         data for GP
+     - Estimate hyperparameters for the GP used to predict the signal shape
+       for every voxel
+     - Update EC and movement parameters
+     - For all i in N (DWIs)
+       - Draw a prediction s_{i} from the GP
+       - Compute \\hat{f}_{i} (s_{i}, h, \beta_{i}, r_{i}, a_{i})
+       - Use \\hat{f}_{i} - f_{i} to update \beta_{i} and r_{i} (eq 6)
```
oesteban, line 445: Estimator does this:
```suggestion
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+    vector defining the PE direction (such that for example [1 0 0], [−1 0 0],
+    [0 1 0] and [0 −1 0] denote R → L, L → R, P → A and A  P PE direction
+    respectively) and where t denotes the readout time (in seconds).
+    r: 1x6 vector: 3 translations, 3 rotations
+    \beta: four for linear; ten for quadratic, twenty for cubic.
+    h: assumed to be in the same space as the first b = 0 image supplied to
+    eddy, which will be automatically fulfilled if it was estimated by topup
+    and that same b = 0 image was the first of those supplied to topup. Hence,
+    it can be said to help define the reference/undistorted space as the first
+    b = 0 image after distortion correction by h.
```
oesteban, line 459: I think this can be moved to some documentation file under doc/

docs/notebooks/dwi_gp_representation.ipynb
oesteban, line 1: Let's move into a separate PR

src/eddymotion/model/base.py
```diff
@@ -127,6 +128,8 @@ def __init__(self, gtab, S0=None, mask=None, b_max=None, **kwargs):
         if not model_str:
             raise TypeError("No model defined")
 
+        # ToDo
+        # Use lazy loading ?
```
oesteban, line 132: Please open an issue to suggest incidental findings/actions
```suggestion
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+        "_r",
+        "_gpr",
+        "_model",
+    )
+
+    def __init__(self, dwi, a, h, kernel, num_iterations=5, **kwargs):
+        """Implement object initialization."""
+
+        self._dwi = dwi  # The HDF5 file object: avoid having the entire 4D volume in memory
+        self._a = a
```
oesteban, line 491: Data are assumed to be corrected for susceptibility (Estimator does this)
```suggestion
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+        self._h = h
+        self._num_iterations = num_iterations
+
+        # Initialize
+        self._betas = 0
+        self._r = 0
+
+        # ToDo
+        # Build the GP kernel here or in fit ?
```
oesteban, line 500: In fit
```suggestion
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+        self._gpr = None
+        # Does the kernel depend on which data we use as the training data (i.e.
+        # varies with the index we choose to predict)?
+        self._kernel = kernel
```
oesteban, line 504: I don't think it does. I believe we just need to keep the kernel between iterations.
```suggestion
        self._kernel = kernel
```

src/eddymotion/model/base.py
```diff
@@ -409,6 +412,119 @@ class DKIModel(BaseModel):
...
+        volumes.
+        X: array-like of shape (n_samples, n_features), n_samples being
+        the number of gradients, and the n_features the number of shells ?
+        Or n_samples being the number of voxels in the DWI volume, and n_features
+        being 3 (bvec coordinates)
+        y: _array-like of shape (n_samples,) or (n_samples, n_targets)"""
+
+        self._gpr = GaussianProcessRegressor(kernel=self._kernel, random_state=0)
+        self._gpr.fit(X, y)
+
```
oesteban, line 518: I think it is here where we update hyperparameters.

src/eddymotion/model/utils.py
oesteban, line 1: I believe all of this is already added or soon be added in alternative PRs, correct? If so, please remove.

test/test_model.py
```diff
@@ -91,6 +92,64 @@ def test_average_model():
...
+                    # ToDo
+                    # Use a mask to avoid traversing background data
+                    y = dwi_data[i, j, k, lovo_idx]
+                    gp.fit(X, y)
+                    prediction, _ = gp.predict(
+                        bvecs[idx, :][np.newaxis]
+                    )  # Can take multiple values X[:2, :]
+                    result[i, j, k, idx] = prediction.item()
+
+    assert result.shape == dwi_data.shape
```
oesteban, line 150: Let's use the notebook for this, for the time being. The only tests to add at this point are:
- The model can be instatiated
- We can fit and predict one voxel - data could be drawn from scikit-learn's examples so that one of the default kernels is used.

```suggestion
```

oesteban (commented):
I finally could have a look into this. I think we have generated some overlap with other efforts, so a first step would be to remove duplicate work (IMHO).

The only major issue I can see is that it looks like the model confuses responsibilities with the estimator at points.

Let's set the following goal to this PR:

Be able to produce one prediction (i.e., one missing orientation) for one voxel (having all the other b-vecs/vals). This should not bother about:
- hyperparameter update
- kernel
- motion and EC parameters update

src/eddymotion/model/base.py
jhlegarreta: Removed in 5d76b88.

src/eddymotion/model/base.py
jhlegarreta: Removed in 5d76b88.

src/eddymotion/model/base.py
jhlegarreta: Removed in 5d76b88.

src/eddymotion/model/base.py
jhlegarreta: Removing all this in 5d76b88. I will document properly when we reach a more solid basis. Also, these do not belong to the GP process (Anderson 2015), strictly but to the correction step (Anderson 2016).

src/eddymotion/model/base.py
jhlegarreta: Same comment as in https://github.com/nipreps/eddymotion/pull/188#discussion_r1632118544.

src/eddymotion/model/base.py
jhlegarreta: Removed all estimator doc in 5d76b88.

src/eddymotion/model/base.py
jhlegarreta: Removed all estimator doc in 5d76b88.

src/eddymotion/model/base.py
jhlegarreta: Will do.

src/eddymotion/model/base.py
jhlegarreta: Removed in 5d76b88.

src/eddymotion/model/base.py
jhlegarreta: Removed in 5d76b88.

src/eddymotion/model/utils.py
jhlegarreta: Yes, removed in 5d76b88.

test/test_model.py
jhlegarreta: Done in 061f130.

jhlegarreta:
Re https://github.com/nipreps/eddymotion/pull/188#pullrequestreview-2104702667 agreed.

Have done a local copy of the old file to add a documentation file as time permits. For now,
- This reduces the whole thing to making the prediction.
- Have removed all duplicates that this had after having split the contents across multiple PRs (some already merged, others being reviewed, etc.)

Will improve things in subsequent commits: actually use the DWI data, use bvec indexing, etc.

src/eddymotion/model/base.py
```diff
@@ -448,6 +449,69 @@ class DKIModel(BaseModel):
...
+        "_gpr",
+    )
+
+    def __init__(self, dwi, kernel, **kwargs):
+        """Implement object initialization.
+
+        Parameters
+        ----------
+        dwi : :obj:`~eddymotion.dmri.DWI`
+            The DWI data.
```
oesteban, line 468: Do we need this dwi argument? It seems it is not used

src/eddymotion/model/base.py
```diff
@@ -448,6 +449,69 @@ class DKIModel(BaseModel):
     _model_class = "dipy.reconst.dki.DiffusionKurtosisModel"
 
 
+class GaussianProcessModel:
+    """A Gaussian process model for DWI data based on [Andersson15]_."""
+
+    __slots__ = (
+        "_dwi",
+        "_kernel",
+        "_num_iterations",
```
oesteban, line 458: What is this slot for? How can it be used at the moment?

src/eddymotion/model/base.py
```diff
@@ -448,6 +449,69 @@ class DKIModel(BaseModel):
     _model_class = "dipy.reconst.dki.DiffusionKurtosisModel"
 
 
+class GaussianProcessModel:
```
oesteban, line 452: Do we want to inherit from model base?
```suggestion
class GaussianProcessModel(ModelBase):
```

oesteban (commented):
looking great! the interface is now very clear

src/eddymotion/model/base.py
oesteban: duh, we need to merge #176

src/eddymotion/model/base.py
jhlegarreta: Removed in 10b4a6b. Sorry.

src/eddymotion/model/base.py
jhlegarreta: Not used yet, but the way I see things we would give a `DWI` instance to the class, which would compute the kernels with the gradient table, and we would give it an index for the fit/predict part. Otherwise, do you think this should be the estimator's task?

src/eddymotion/model/base.py
jhlegarreta: PR is merged. Done in 10b4a6b.

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+        X : :obj:`~numpy.ndarray` of shape (n_samples,)
+            Query points where the Gaussian process is evaluated: the
+            diffusion-encoding gradient vectors of interest.
+
+        Returns
+        -------
+        y_mean : :obj:`~numpy.ndarray` of shape (n_samples,) or (n_samples, n_targets)
+            Mean of predictive distribution at query points.
+        y_std : :obj:`~numpy.ndarray` of shape (n_samples,) or (n_samples, n_targets)
+            Standard deviation of predictive distribution at query points.
```
oesteban, line 581: This needs to be the simulated volume corresponding to the input gradient.

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+    def predict(self, X, **kwargs):
+        """Predict using the Gaussian process model of the DWI signal, where
+        ``X`` is a diffusion-encoding gradient vector whose DWI data needs to be
+        estimated.
+
+        Parameters
+        ----------
+        X : :obj:`~numpy.ndarray` of shape (n_samples,)
+            Query points where the Gaussian process is evaluated: the
+            diffusion-encoding gradient vectors of interest.
```
oesteban, line 574:
```suggestion
        X : :obj:`~numpy.ndarray` of shape (n_dimensions,)
            A diffusion-encoding gradient vector for which we want to predict signal.
```

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+        ----------
+        X : :obj:`~numpy.ndarray` of shape (n_samples, n_features)
+            Feature values for training. For the DWI cae, ``n_samples`` is the
+            number of diffusion-encoding gradient vectors, and ``n_features``
+            being 3 (the spatial coordinates).
+        y : :obj:`~numpy.ndarray` of shape (n_samples,) or (n_samples, n_targets)
+            Target values: the DWI signal values.
+        """
+
+        self._gpr = GaussianProcessRegressor(kernel=self._kernel, random_state=0)
```
oesteban, line 561: Should not `kernel` be an object (as opposed to a type) here?
```suggestion
        self._gpr = GaussianProcessRegressor(kernel=self._kernel(X), random_state=0)
```

test/test_model.py
```diff
@@ -105,6 +107,23 @@ def test_average_model():
     assert np.all(tmodel_2000.predict([0, 0, 0]) == 1100)
 
 
+def test_gp_model(datadir):
+    dwi = DWI.from_filename(datadir / "dwi.h5")
+
+    kernel = DotProduct() + WhiteKernel()
```
oesteban, line 113: It does look like kernel is an object not a type.

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+class GaussianProcessModel(BaseModel):
+    """A Gaussian process model for DWI data based on [Andersson15]_."""
+
+    __slots__ = (
+        "_dwi",
+        "_kernel",
+        "_gpr",
+    )
+
+    def __init__(self, dwi, kernel, **kwargs):
```
oesteban, line 534: I think we can initalize with gtab and avoid dragging the dwi data within the object.
```suggestion
    def __init__(self, gtab, kernel, **kwargs):
```

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
     _model_class = "dipy.reconst.dki.DiffusionKurtosisModel"
 
 
+class GaussianProcessModel(BaseModel):
+    """A Gaussian process model for DWI data based on [Andersson15]_."""
+
+    __slots__ = (
+        "_dwi",
```
oesteban, line 529:
```suggestion
        "_gtab",
```

src/eddymotion/model/base.py
oesteban: Ha, no I see we had already discussed this. I think this is only necessary in fit()

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+
+        Parameters
+        ----------
+        dwi : :obj:`~eddymotion.dmri.DWI`
+            The DWI data.
+        kernel : :obj:`~sklearn.gaussian_process.kernels.Kernel`
+            Kernel instance.
+        """
+
+        self._dwi = dwi
```
oesteban, line 545:
```suggestion
        self._gtab = _rasb2dipy(gtab)
```

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+        self._kernel = kernel
+
+    def fit(self, X, y, *args, **kwargs):
```
oesteban, line 548:
```suggestion
    def fit(self, gtab, data, *args, **kwargs):
```

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+        X : :obj:`~numpy.ndarray` of shape (n_samples, n_features)
+            Feature values for training. For the DWI cae, ``n_samples`` is the
+            number of diffusion-encoding gradient vectors, and ``n_features``
+            being 3 (the spatial coordinates).
+        y : :obj:`~numpy.ndarray` of shape (n_samples,) or (n_samples, n_targets)
+            Target values: the DWI signal values.
+        """
+
+        self._gpr = GaussianProcessRegressor(kernel=self._kernel, random_state=0)
+        self._gpr.fit(X, y)
```
oesteban, line 562:
```suggestion
        self._gpr.fit(gtab, data)
```

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+        self._is_fitted = True
+
+    def predict(self, X, **kwargs):
```
oesteban, line 565:
```suggestion
    def predict(self, gradient, **kwargs):
```

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+        y_mean : :obj:`~numpy.ndarray` of shape (n_samples,) or (n_samples, n_targets)
+            Mean of predictive distribution at query points.
+        y_std : :obj:`~numpy.ndarray` of shape (n_samples,) or (n_samples, n_targets)
+            Standard deviation of predictive distribution at query points.
+        """
+
+        if not self._is_fitted:
+            raise ModelNotFittedError(f"{type(self).__name__} must be fitted before predicting")
+
+        y_mean, y_std = self._gpr.predict(X, return_std=True)
```
oesteban, line 587:
```suggestion
        y_mean, y_std = self._gpr.predict(gradient, return_std=True)
```

src/eddymotion/model/base.py
```diff
@@ -521,6 +522,72 @@ class DKIModel(BaseDWIModel):
...
+        return y_mean, y_std
```
oesteban, line 588: massage so it is a 3D volume

```suggestion
        return y_mean, y_std
```

test/test_model.py
jhlegarreta: According to the scikit-learn example,
https://scikit-learn.org/stable/auto_examples/gaussian_process/plot_gpr_noisy_targets.html#example-with-noise-free-target

It is an object, yes, so not sure what you want to change in here

src/eddymotion/model/base.py
jhlegarreta: `X` should correspond to our `gtab` and `y` to our `data`. I see that the first version of the comment followed that rationale. Discussing this over a call would be best.

src/eddymotion/model/base.py
jhlegarreta: Contradicts https://github.com/nipreps/eddymotion/pull/188#discussion_r1640211158.

src/eddymotion/model/base.py
jhlegarreta: Related to https://github.com/nipreps/eddymotion/pull/188#discussion_r1641433635.

jhlegarreta:
Have gone through the comments but I am hesitant about them:
- Now I do see that we should not provide the diffusion signal to the `__init__` method. At least not at this stage of the development.
- Re https://github.com/nipreps/eddymotion/pull/188#discussion_r1638260136 https://github.com/nipreps/eddymotion/pull/188#discussion_r1640215088 I think here we are back to the discussion about our data: reconstructing the 3D volume in here means that the fit/predict method get the entire data, which is unfeasible given the size. As I see things, reconstructing the 3D volume should be done in the caller. Best would be to discuss this over a call.

oesteban:
> Now I do see that we should not provide the diffusion signal to the `__init__` method. At least not at this stage of the development.

👍 

> Re https://github.com/nipreps/eddymotion/pull/188#discussion_r1638260136 https://github.com/nipreps/eddymotion/pull/188#discussion_r1640215088 I think here we are back to the discussion about our data: reconstructing the 3D volume in here means that the fit/predict method get the entire data, which is unfeasible given the size. As I see things, reconstructing the 3D volume should be done in the caller. Best would be to discuss this over a call.

Happy to have a call, but IMHO, this is an 'eddymotion' model -- meaning, we define the interface. All other models generate volumes as the output of their predict(), so this needs to be another one. Perhaps what's missing here is an 'intermediate' model (which could eventually go into DIPY, cc/ @arokem) where the output would be consistent with the standard GP output (i.e., a mean and std functions).

src/eddymotion/model/base.py
oesteban: I agree, if we want to be consistent with scikit-learn. However, this will require us to change all other methods' signature, I think.

jhlegarreta:
> Happy to have a call, but IMHO, this is an 'eddymotion' model -- meaning, we define the interface. All other models generate volumes as the output of their predict(), so this needs to be another one. Perhaps what's missing here is an 'intermediate' model (which could eventually go into DIPY, cc/ @arokem) where the output would be consistent with the standard GP output (i.e., a mean and std functions).

OK. I see that. So the input will need to be the entire DWI image then. We'll then need to figure out how this can be dealt with within the model. DIPY may give us clues, yes. For now, I do not see how/have not digged into it.

It can still be tested in its current form (without additional strategies, running into memory issues, etc.) with https://github.com/nipreps/eddymotion/blob/a8742d955d9a43676d54ef194737a42db6f74015/docs/notebooks/dwi_simulated_gp.ipynb

arokem:
On Mon, Jun 17, 2024 at 5:23 AM Jon Haitz Legarreta Gorroño <
***@***.***> wrote:

> Happy to have a call, but IMHO, this is an 'eddymotion' model -- meaning,
> we define the interface. All other models generate volumes as the output of
> their predict(), so this needs to be another one. Perhaps what's missing
> here is an 'intermediate' model (which could eventually go into DIPY, cc/
> @arokem <https://github.com/arokem>) where the output would be consistent
> with the standard GP output (i.e., a mean and std functions).
>
> OK. I see that. So the input will need to be the entire DWI image then.
> We'll then need to figure out how this can be dealt with within the model.
> DIPY may give us clues, yes. For now, I do not see how/have not digged into
> it.
>

If you are implementing a new DIPY-style model here it would be best to
inherit from ‘ReconstModel’ and ‘ReconstFit’, which can take care of some
of the dispatch and will take care of parallelization once
https://github.com/dipy/dipy/pull/2593 is merged.

src/eddymotion/model/base.py
jhlegarreta: This is now deprecated after https://github.com/jhlegarreta/eddymotion/pull/7, neither the signal nor the gradient table are slots of the class in the new signature, so resolving. We'll see how the new framework comes into life after we stabilize things.

src/eddymotion/model/base.py
jhlegarreta: This is now deprecated after https://github.com/jhlegarreta/eddymotion/pull/7, neither the signal nor the gradient table are slots of the class in the new signature, so resolving. We'll see how the new framework comes into life after we stabilize things.

jhlegarreta: This is now deprecated after https://github.com/jhlegarreta/eddymotion/pull/7, neither the signal nor the gradient table are provided when initializing the class in the new signature, so resolving. We'll see how the new framework comes into life after we stabilize things.

src/eddymotion/model/base.py
jhlegarreta: I now realize that this switched the order to provides the parameters in the ordering that scikit-learn requires, so resolving.

src/eddymotion/model/base.py
jhlegarreta: OK, done in https://github.com/jhlegarreta/eddymotion/pull/7, so resolving.

src/eddymotion/model/base.py
jhlegarreta: After https://github.com/jhlegarreta/eddymotion/pull/7 the implementation calling it `gradient` has prevailed, so resolving.

src/eddymotion/model/base.py
jhlegarreta: The implementation in https://github.com/jhlegarreta/eddymotion/pull/7 adopts a gradient table instance, closer to the suggested description, so resolving.

src/eddymotion/model/base.py
jhlegarreta: :+1:. I used a naming that was closer to what scikit-learn uses so that the exercise of identifying our data with what it uses was done at an earlier stage, but resolving now as the naming adopted in https://github.com/jhlegarreta/eddymotion/pull/7 is closer to the dMRI realm.

jhlegarreta:
@oesteban Please check the following in commit 5bdff10:
- https://github.com/nipreps/eddymotion/pull/188/commits/5bdff10782e123a34927b41329d3a23c2d4691ab#diff-fba3c6e90b496595a3cd3cdcbd5d562082cde0b4f21629d667d1a2260a91bd37L67
- https://github.com/nipreps/eddymotion/pull/188/commits/5bdff10782e123a34927b41329d3a23c2d4691ab#diff-fba3c6e90b496595a3cd3cdcbd5d562082cde0b4f21629d667d1a2260a91bd37R191

I just made it such that the test passes, but it may not fit into what you had envisioned (especially the latter, which will probably fail when giving it real data/using a mask).

test/test_model.py
```diff
@@ -105,6 +108,22 @@ def test_average_model():
     assert np.all(tmodel_2000.predict([0, 0, 0]) == 1100)
 
 
+def test_gp_model():
+    gp = GaussianProcessModel("test", kernel="default")
```
oesteban, line 112:
```suggestion
    gp = GaussianProcessModel("test")
```

oesteban (approved):
Good stuff - merging as soon as tests get green, thanks for this!

src/eddymotion/model/dipy.py
```diff
@@ -0,0 +1,419 @@
...
+    """
+
+    # Check it's fitted as they do in sklearn internally
+    # https://github.com/scikit-learn/scikit-learn/blob/972e17fe1aa12d481b120ad4a3dc076bae736931/\
+    # sklearn/gaussian_process/_gpr.py#L410C9-L410C42
+    if not hasattr(model, "X_train_"):
+        raise RuntimeError("Model is not yet fitted.")
+
+    # Extract orientations from gtab, and highly likely, the b-value too.
+    return model.predict(gtab, return_std=False)
```
jhlegarreta, line 71: I am trying to debug this  
https://github.com/nipreps/eddymotion/pull/228#discussion_r1776148824

While at it, I see that `model_gtab` is not used by this method. Although I may imagine what was the underlying intention (much what we try to do in https://github.com/nipreps/eddymotion/pull/228#discussion_r1776148824), not sure how you intended to implement this. This is called from:
https://github.com/nipreps/eddymotion/pull/188/files#diff-fba3c6e90b496595a3cd3cdcbd5d562082cde0b4f21629d667d1a2260a91bd37R281

The mask is not yet used, and although we have discussed using the mask in the past, it should probably be related to the use of `gtab` and `model_gtab`.

@oesteban any thoughts about this?

PR #189, opened 2024-05-06
ENH: Call the DWI plot method from nireports

jhlegarreta:
Call the DWI plot method from its new location in `nireports`.


jhlegarreta:
Depends on https://github.com/nipreps/nireports/pull/101 and a new release of `nireports`.

oesteban (approved):
LGTM - thanks!

oesteban:
Please change from draft to "ready for review" if it is ready.

jhlegarreta:
OK, I see that `nireports` is not pinned, and since a new release was made with this, this is ready to go.

PR #198, opened 2024-05-27
ENH: Add DWI shell detection utils

jhlegarreta:
Add DWI shell detection utils.

Add the corresponding tests.


jhlegarreta:
@oesteban We may need this at some point. Ready to be merged.

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        if kmeans_res.inertia_ / len(bvals) < tol:
+            break
+    else:
+        raise ValueError(f"bvals parsing failed: no shells found more than {tol} apart")
+
+    # Convert the kclust labels to an array
+    shells = kmeans_res.cluster_centers_
+    bval_centroids = np.zeros(bvals.shape)
+    for i in range(shells.size):
+        bval_centroids[kmeans_res.labels_ == i] = shells[i][0]
```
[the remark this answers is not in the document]
jhlegarreta: Sorry, was not aware of that code snippet. If that does a better job than the one proposed here, happy to change. I have tried it locally and looks to perform well.

> That said, perhaps a more straightforward way of doing this is to just bin the b-values into, say, 15 bins with a range of b-values from 0 through, say, 8,000. Then, pick nonempty bins, and if they are below seven, call the dataset multi-shell, and if they are above seven, call them DSI.

Not sure if I see this: are you saying that the `mrqc` snippet should be edited (assuming we C&P here) following what you are describing here? Do you think that the current `mriqc` implementation will be more robust with those changes? Unless the `mriqc` implementation is flawed, I would leave that for a separate PR. We can open an issue once this is merged.

src/eddymotion/model/tests/test_dmri_utils.py
```diff
@@ -0,0 +1,261 @@
...
+    tol = 20
+    bvals = np.asarray([0, 0])
+    expected_shells = np.asarray([0])
+    expected_bval_centroids = np.asarray([0, 0])
+    obtained_shells, obtained_bval_centroids = find_shelling_scheme(bvals, tol=tol)
+
+    assert np.array_equal(obtained_shells, expected_shells)
+    assert np.array_equal(obtained_bval_centroids, expected_bval_centroids)
+
+    bvals = np.asarray(
```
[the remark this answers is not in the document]
jhlegarreta: The DSI dataset was added directly to nireports:
https://github.com/nipreps/nireports/pull/96/files

Not sure if it exists elsewhere.

Here, we only point to an HDF5 file downloaded from OSF (https://github.com/nipreps/eddymotion/blame/2b07b0e0e0e34b2938b9a1d34b795d3037072002/docs/usage.rst#L27) or gin.g-node (https://github.com/nipreps/eddymotion/blob/2b07b0e0e0e34b2938b9a1d34b795d3037072002/.circleci/config.yml#L62).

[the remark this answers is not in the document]
oesteban: > The DSI dataset was added directly to nireports: https://github.com/nipreps/nireports/pull/96/files
> 
> Not sure if it exists elsewhere.

Right, it's nireports - sorry for the confusion. Let's add those bvec/vals files here right?

src/eddymotion/model/dmri_utils.py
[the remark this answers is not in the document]
oesteban: > Sorry, was not aware of that code snippet. If that does a better job than the one proposed here, happy to change. I have tried it locally and looks to perform well.

What works well? The mriqc snippet, this snippet or both :D



> Not sure if I see this: are you saying that the `mrqc` snippet should be edited (assuming we C&P here) following what you are describing here?

Yes, my impression is that k-means is an overkill for this problem. Just calculating the histogram with the adequate number of bins and min/max definitions should suffice.

So, instead of C&P, I guess we could try the solution here en eddymotion and later propagate into mriqc if it is successful.

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,77 @@
...
+"""Minimum allowed value for the b-value."""
+
+DEFAULT_NUM_BINS = 15
+"""Number of bins to classify b-values."""
+
+DEFAULT_MULTISHELL_BIN_COUNT_THR = 7
+"""Default bin count to consider a multishell scheme."""
+
+
+def find_shelling_scheme(
```
oesteban, line 35: this is pretty much what I had in mind 👍 

If it works, then it works!

oesteban (approved):

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,77 @@
...
+    bvals,
+    num_bins=DEFAULT_NUM_BINS,
+    multishell_nonempty_bin_count_thr=DEFAULT_MULTISHELL_BIN_COUNT_THR,
+):
+    """Find the shelling scheme on the given b-values: computes the histogram of
+    the b-values according to ``num_bins`` and depending on the nonempty bin
+    count, classify the shelling scheme as: single-shell if they are 2;
+    multi-shell if they are below the ``multishell_nonempty_bin_count_thr``
+    value; DSI otherwise.
```
oesteban, line 44:
```suggestion
    """
    Find the shelling scheme on the given b-values.
    
    Computes the histogram of the b-values according to ``num_bins``
    and depending on the nonempty bin count, classify the shelling scheme 
    as single-shell if they are 2 (low-b and a shell); multi-shell if they are below
    the ``multishell_nonempty_bin_count_thr`` value; and DSI otherwise.
```

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Parameters
+    ----------
+    bvals : :obj:`~numpy.ndarray`
+         b-values in FSL format.
```
oesteban, line 49:
```suggestion
         List or array of b-values.
```

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,77 @@
...
+):
+    """Find the shelling scheme on the given b-values: computes the histogram of
+    the b-values according to ``num_bins`` and depending on the nonempty bin
+    count, classify the shelling scheme as: single-shell if they are 2;
+    multi-shell if they are below the ``multishell_nonempty_bin_count_thr``
+    value; DSI otherwise.
+
+    Parameters
+    ----------
+    bvals : :obj:`~numpy.ndarray`
```
oesteban, line 48:
```suggestion
    bvals : :obj:`list` or :obj:`~numpy.ndarray`
```

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,77 @@
...
+    nonempty_bins = np.sum(hist > 0)
+
+    if nonempty_bins == 2:
+        scheme = "single-shell"
+    elif nonempty_bins < multishell_nonempty_bin_count_thr:
+        scheme = "multi-shell"
+    else:
+        scheme = "DSI"
+
+    return scheme, hist, bin_edges
```
oesteban, line 77: Should we just return `scheme` and the actual "shells"? For example, estimated as the median value of all the b-values in each nonempty bin.

oesteban (approved):

src/eddymotion/model/dmri_utils.py
jhlegarreta: Done in b41a427.

src/eddymotion/model/dmri_utils.py
jhlegarreta: Done in b41a427.

src/eddymotion/model/dmri_utils.py
jhlegarreta: Done in b41a427.

src/eddymotion/model/tests/test_dmri_utils.py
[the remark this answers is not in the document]
jhlegarreta: Added.

src/eddymotion/model/dmri_utils.py
jhlegarreta: Reworked in b41a427 to return the actual values in each bin. Overall, not sure if the histogram approach is better or more robust than the k-means, but we'll find out as the tool is used. Resolving now. Ready to be merged.

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    bins = np.linspace(-1, max(bvals), num_bins + 1)
+    hist, bin_edges = np.histogram(bvals, bins=bins)
+
+    # Collect values in each bin
+    bval_groups = []
+    for lower, upper in zip(bin_edges[:-1], bin_edges[1:], strict=False):
+        bval_groups.append(bvals[(bvals > lower) & (bvals <= upper)])
+
+    # Remove empty bins from the list
+    bval_groups = [v for v in bval_groups if len(v)]
```
oesteban, line 73: The reason why I suggested taking the b-values in each bin and calculating the median (I may have said mean) is that the bin edges are artificial: if you have bin edges falling in, e.g., 1000-1100, a single shell DTI with b-value = 1000 will report here 1050 as the b-value.

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    -------
+    scheme : :obj:`str`
+        Shelling scheme.
+    bval_groups : :obj:`list`
+        List of grouped b-values.
+    """
+
+    # Bin the b-values: use -1 as the lower bound to be able to appropriately
+    # include b0 values
+    bins = np.linspace(-1, max(bvals), num_bins + 1)
```
oesteban, line 64:
```suggestion
    bvals = np.array(bvals)
    bins = np.linspace(-1, max(bvals.max(), DEFAULT_MAX_BVAL), num_bins + 1)
```
I believe we want to set a large maximum bval - if the expected b-values were not reached, we may have a dataset with, e.g., 920, 1000, 950, 990, 1000, 950 and if the maximum is 1000 with 15 bins, then 920 will probably fall in a different bin.

oesteban: with `DEFAULT_MAX_BVAL=10000` or something of the sort.

src/eddymotion/model/dmri_utils.py
jhlegarreta: You said median, that was correct. So do you want to add the median as e.g. a key to a dict where values are the actual b-values? I am not returning the mean; I am returning the b-values from the `bvals` input. Maybe easier if you directly amend the commit with what you have in mind.

src/eddymotion/model/dmri_utils.py
jhlegarreta: If you are going to amend the commit following https://github.com/nipreps/eddymotion/pull/198#discussion_r1632078801, maybe you can modify this as well.

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,84 @@
...
+#     https://www.nipreps.org/community/licensing/
+#
+import numpy as np
+
+DEFAULT_NUM_BINS = 15
+"""Number of bins to classify b-values."""
+
+DEFAULT_MULTISHELL_BIN_COUNT_THR = 7
+"""Default bin count to consider a multishell scheme."""
+
```
oesteban, line 30:
```suggestion

DEFAULT_MAX_BVAL = 8000
"""Maximum b-value cap."""
```

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    scheme : :obj:`str`
+        Shelling scheme.
+    bval_groups : :obj:`list`
+        List of grouped b-values.
+    """
+
+    # Bin the b-values: use -1 as the lower bound to be able to appropriately
+    # include b0 values
+    bins = np.linspace(-1, max(bvals), num_bins + 1)
+    hist, bin_edges = np.histogram(bvals, bins=bins)
```
oesteban, line 65:
```suggestion
    hist, bin_edges = np.histogram(bvals, bins=num_bins, range=(-1, min(max(bvals), bval_cap)))
```

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,84 @@
...
+"""Number of bins to classify b-values."""
+
+DEFAULT_MULTISHELL_BIN_COUNT_THR = 7
+"""Default bin count to consider a multishell scheme."""
+
+
+def find_shelling_scheme(
+    bvals,
+    num_bins=DEFAULT_NUM_BINS,
+    multishell_nonempty_bin_count_thr=DEFAULT_MULTISHELL_BIN_COUNT_THR,
```
oesteban, line 35:
```suggestion
    multishell_nonempty_bin_count_thr=DEFAULT_MULTISHELL_BIN_COUNT_THR,
    bval_cap=DEFAULT_MAX_BVAL,
```

src/eddymotion/model/dmri_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        Number of bins.
+    multishell_nonempty_bin_count_thr : :obj:`int`, optional
+        Bin count to consider a multi-shell scheme.
+
+    Returns
+    -------
+    scheme : :obj:`str`
+        Shelling scheme.
+    bval_groups : :obj:`list`
+        List of grouped b-values.
```
oesteban, line 59:
```suggestion
        List of grouped b-values.
    bval_estimated : :obj:`list`
        List of 'estimated' b-values as the median value of each b-value group.

```

src/eddymotion/model/tests/test_dmri_utils.py
```diff
@@ -0,0 +1,620 @@
...
+                    ]
+                ),
+            ],
+        ),
+    ],
+)
+def test_find_shelling_scheme_files(dwi_btable, exp_scheme, exp_bval_groups):
+    bvals = np.loadtxt(_datadir / f"{dwi_btable}.bval")
+
+    obt_scheme, obt_bval_groups = find_shelling_scheme(bvals)
```
oesteban, line 615:
```suggestion
    obt_scheme, obt_bval_groups, obt_bval_estimates = find_shelling_scheme(bvals)
```

jhlegarreta: Left behind in my previous commit. Added in d221f03.

oesteban (approved):

oesteban:
I think this is ready! thanks a ton! Please merge if you think all is okay :)

PR #200, opened 2024-05-27
ENH: Add gradient encoding direction angle computation utils

jhlegarreta:
Add gradient encoding direction angle computation utils: compute pairwise angles across gradient encoding directions.

Add the corresponding test.

### DIFF
--- a/src/eddymotion/math/utils.py
+++ b/src/eddymotion/math/utils.py
@@ -29,7 +29,7 @@ def is_positive_definite(matrix):
     matrix and its conjugate transpose by performing the Cholesky decomposition.
     Parameters
     ----------
-    matrix : np.ndarray
+    matrix : :obj:`~numpy.ndarray`
         The matrix to check.
     Returns
     -------
--- a/src/eddymotion/math/utils.py
+++ b/src/eddymotion/math/utils.py
@@ -43,3 +43,47 @@ def is_positive_definite(matrix):
     except np.linalg.LinAlgError:
         # Matrix is not positive definite
         return False
+
+
+def compute_angle(v1, v2, closest_polarity=False):
+    """Compute the angle between two vectors.
+
+    Parameters
+    ----------
+    v1 : :obj:`~numpy.ndarray`
+        First vector.
+    v2 : :obj:`~numpy.ndarray`
+        Second vector.
+    closest_polarity : :obj:`bool`
+        ``True`` to consider the smallest of the two angles between the crossing
+         lines resulting from reversing both vectors.
+
+    Returns
+    -------
+    :obj:`float`
+        The angle between the two vectors in radians.
+
+    Examples
+    --------
+    >>> compute_angle(
+    ...     np.array((1.0, 0.0, 0.0)),
+    ...     np.array((-1.0, 0.0, 0.0)),
+    ... )  # doctest: +ELLIPSIS
+    3.1415...
+    >>> compute_angle(
+    ...     np.array((1.0, 0.0, 0.0)),
+    ...     np.array((-1.0, 0.0, 0.0)),
+    ...     closest_polarity=True,
+    ... )
+    0.0
+
+    """
+
+    cosine_angle = (v1 / np.linalg.norm(v1)) @ (v2 / np.linalg.norm(v2))
+    # Clip values to handle numerical errors
+    cosine_angle = np.clip(
+        np.abs(cosine_angle) if closest_polarity else cosine_angle,
+        -1.0,
+        1.0,
+    )
+    return np.arccos(cosine_angle)
--- a/src/eddymotion/model/gradient_utils.py
+++ b/src/eddymotion/model/gradient_utils.py
@@ -1,0 +1,75 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2024 The NiPreps Developers <nipreps@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+import numpy as np
+
+
+def compute_pairwise_angles(bvecs, closest_polarity):
+    r"""Compute pairwise angles across diffusion gradient encoding directions.
+
+    Following [Andersson15]_, it computes the smallest of the angles between
+    each pair if ``closest_polarity`` is ``True``, i.e.
+
+    .. math::
+
+        \theta(\mathbf{g}, \mathbf{g'}) = \arccos(\abs{\langle \mathbf{g}, \mathbf{g'} \rangle})
+
+    Parameters
+    ----------
+    bvecs : :obj:`~numpy.ndarray`
+        Diffusion gradient encoding directions in FSL format.
+    closest_polarity : :obj:`bool`
+        ``True`` to consider the smallest of the two angles between the crossing
+         lines resulting from reversing each vector pair.
+
+    Returns
+    -------
+    :obj:`~numpy.ndarray`
+        Pairwise angles across diffusion gradient encoding directions.
+
+    Examples
+    --------
+    >>> compute_pairwise_angles(
+    ...     ((1.0, -1.0), (0.0, 0.0), (0.0, 0.0)),
+    ...     False,
+    ... )[0, 1]  # doctest: +ELLIPSIS
+    3.1415...
+    >>> compute_pairwise_angles(
+    ...     ((1.0, -1.0), (0.0, 0.0), (0.0, 0.0)),
+    ...     True,
+    ... )[0, 1]
+    0.0
+
+    References
+    ----------
+    .. [Andersson15] J. L. R. Andersson. et al., An integrated approach to
+       correction for off-resonance effects and subject movement in diffusion MR
+       imaging, NeuroImage 125 (2016) 1063–1078
+    """
+
+    if np.shape(bvecs)[0] != 3:
+        raise ValueError(f"bvecs must be of shape (3, N). Found: {bvecs.shape}")
+
+    # Ensure b-vectors are unit-norm
+    bvecs = np.array(bvecs) / np.linalg.norm(bvecs, axis=0)
+    cosines = np.clip(bvecs.T @ bvecs, -1.0, 1.0)
+    return np.arccos(np.abs(cosines) if closest_polarity else cosines)
--- a/src/eddymotion/math/tests/test_utils.py
+++ b/src/eddymotion/math/tests/test_utils.py
@@ -21,8 +21,9 @@
 #     https://www.nipreps.org/community/licensing/
 #
 import numpy as np
+import pytest
 
-from eddymotion.math.utils import is_positive_definite
+from eddymotion.math.utils import compute_angle, is_positive_definite
 
 
 def test_is_positive_definite():
@@ -31,3 +32,17 @@ def test_is_positive_definite():
 
     matrix = np.array([[4, 1, 2], [1, -3, 1], [2, 1, 5]])
     assert not is_positive_definite(matrix)
+
+
+@pytest.mark.parametrize(
+    ("v1", "v2", "closest_polarity", "expected"),
+    [
+        ([1, 0, 0], [1, 0, 0], False, 0),
+        ([1, 0, 0], [0, 1, 0], False, np.pi / 2),
+        ([1, 0, 0], [-1 / np.sqrt(2), 0, 1 / np.sqrt(2)], False, np.pi * 3 / 4),
+        ([1, 0, 0], [-1 / np.sqrt(2), 0, 1 / np.sqrt(2)], True, np.pi / 4),
+    ],
+)
+def test_compute_angle(v1, v2, closest_polarity, expected):
+    obtained = compute_angle(v1, v2, closest_polarity=closest_polarity)
+    assert np.isclose(obtained, expected)
--- a/src/eddymotion/model/tests/test_gradient_utils.py
+++ b/src/eddymotion/model/tests/test_gradient_utils.py
@@ -1,0 +1,62 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2024 The NiPreps Developers <nipreps@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+import numpy as np
+
+from eddymotion.model.gradient_utils import compute_pairwise_angles
+
+
+def test_compute_pairwise_angles():
+    # No need to use normalized vectors: compute_angle takes care of dealing
+    # with it.
+    # The last vector serves as a case where e.g. the angle between the first
+    # vector and the last one is 135, and the method yielding the smallest
+    # resulting angle between the crossing lines (45 vs 135)
+    bvecs = np.array(
+        [
+            [1, 0, 0, 1, 1, 0, -1],
+            [0, 1, 0, 1, 0, 1, 0],
+            [0, 0, 1, 0, 1, 1, 1],
+        ]
+    )
+
+    expected = np.array(
+        [
+            [0.0, np.pi / 2, np.pi / 2, np.pi / 4, np.pi / 4, np.pi / 2, np.pi / 4],
+            [np.pi / 2, 0.0, np.pi / 2, np.pi / 4, np.pi / 2, np.pi / 4, np.pi / 2],
+            [np.pi / 2, np.pi / 2, 0.0, np.pi / 2, np.pi / 4, np.pi / 4, np.pi / 4],
+            [np.pi / 4, np.pi / 4, np.pi / 2, 0.0, np.pi / 3, np.pi / 3, np.pi / 3],
+            [np.pi / 4, np.pi / 2, np.pi / 4, np.pi / 3, 0.0, np.pi / 3, np.pi / 2],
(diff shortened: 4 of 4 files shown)


jhlegarreta:
Need to go through https://github.com/nipreps/eddymotion/pull/188#issuecomment-2118215168 and incorporate https://github.com/jhlegarreta/eddymotion/pull/1.

src/eddymotion/model/utils.py
```diff
@@ -0,0 +1,43 @@
...
+    # Should the b0 be included?
+    # What if whe have different bvals ?
+    # Note that the double loops assume bvecs is a 2D array
+
+    n = len(bvecs)
+    angle_mat = np.zeros((n, n))
+
+    for j in range(n):
+        for i in range(j, n):
+            angle_mat[i, j] = np.arccos(min(1.0, abs(np.dot(bvecs[i], bvecs[j]))))
```
arokem, line 41: Im pretty sure you can get rid of the loop and do 'bvecs.T @ bvecs' to get the outer product of all the bvecs in one fell swoop

src/eddymotion/model/utils.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+import numpy as np
+
+
+def calculate_angle_matrix(bvecs):
```
esavary, line 26: I think we should maybe allow the method to accept two different bvecs, e.g., g and g', in case we need to evaluate this between training and unseen data, such as for evaluating k(x*,x) in equation (4) of Andersson 2015.

src/eddymotion/model/tests/test_utils.py
```diff
@@ -0,0 +1,30 @@
...
+#
+#     https://www.nipreps.org/community/licensing/
+#
+import numpy as np
+from eddymotion.model.utils import (
+    calculate_angle_matrix,
+)
+
+
+def test_calculate_angle_matrix():
```
esavary, line 29: I tested the angle computations while working on the hyperparameter optimization. I'm not sure if this aligns with the tests you wanted to do, but I''m dropping my code it here in case it's useful (feel free to ignore if it's not).
```
def test_calculate_angle_matrix():
    # Define test gradient vectors
    gradients = np.array([
        [1, 0, 0],
        [0, 1, 0],
        [0, 0, 1]
    ])
    
    # Calculate the angle matrix
    angle_matrix = calculate_angle_matrix(gradients)
    
    # Expected angle matrix
    expected_angles = np.array([
        [0, np.pi / 2, np.pi / 2],
        [np.pi / 2, 0, np.pi / 2],
        [np.pi / 2, np.pi / 2, 0]
    ])
    
    # Assert the shape of the angle matrix
    assert angle_matrix.shape == expected_angles.shape
    
    # Assert the calculated angles are almost equal to the expected angles
    np.testing.assert_array_almost_equal(angle_matrix, expected_angles, decimal=6)

if __name__ == "__main__":
    pytest.main()
```

src/eddymotion/model/utils.py
jhlegarreta: Agreed Ariel, but for now I would prefer to see that things give what we expect, despite the code not being pythonic and being inefficient if that allows us to read more easily.

src/eddymotion/model/utils.py
jhlegarreta: Resolving for now.

src/eddymotion/model/tests/test_utils.py
jhlegarreta: Thanks Elodie. I add that to my last push force, and incorporated other cases for the array to be rectangular.

jhlegarreta:
Went over @yibeichan's related https://github.com/nipreps/eddymotion/pull/188#issuecomment-2118215168. For now, assuming it is an ndarray, so no need for the checks.

src/eddymotion/model/utils.py
jhlegarreta: Resolving for now Ariel. We'll improve efficiency in a separate PR.

jhlegarreta:
@oesteban ready to be merged.

jhlegarreta:
@esavary would be great if you could have a quick look. Thanks.

esavary (approved):

esavary:
> @esavary would be great if you could have a quick look. Thanks.

I made a few comments on your questions inside the code, but it looks ready to merge to me.

src/eddymotion/model/gradient_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    if bvecs.shape[0] != 3:
+        raise ValueError(f"bvecs must be of shape (3, N). Found: {bvecs.shape}")
+
+    num_vectors = bvecs.shape[1]
+    angles = np.zeros((num_vectors, num_vectors))
+
+    # ToDo
+    # Note that the below also computes the angle of a given vector with itself:
```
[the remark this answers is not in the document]
jhlegarreta: :+1: Keeping it for now then. Resolving.

src/eddymotion/model/gradient_utils.py
```diff
@@ -0,0 +1,81 @@
...
+import numpy as np
+
+from eddymotion.math.utils import compute_angle
+
+
+# ToDo
+# FSL crew takes as input dpars, which are diffusion parameters
+# Does this deal with multi-shell? What does the matrix look like cols vs rows?
+# Should the b0 be included or explicitly mention that it is excluded?
+# What if whe have different bvals ?
```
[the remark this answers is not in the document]
jhlegarreta: Agreed. Let's keep it simple for now and focus on the single-shell case. We will generalize later for the multi-shell case. Resolving.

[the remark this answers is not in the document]
esavary: > Agreed. Let's keep it simple for now and focus on the single-shell case. We will generalize later for the multi-shell case. Resolving.

In fact, I just realized we might already need to generalize the method to two different bvecs for the one-shell case. (I'm currently modifying your notebook to test the code with the custom kernel I built in PR https://github.com/jhlegarreta/eddymotion/pull/2.) If my understanding is correct, we need to feed the predict method with all the angles between the directions of the training set and the leftover direction. I'm running some tests and will keep you updated when I have more informations.

src/eddymotion/math/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    """Compute the angle between two vectors.
+
+    Parameters
+    ----------
+    v1 : :obj:`ndarray`
+        First vector.
+    v2 : :obj:`ndarray`
+        Second vector.
+    smallest : :obj:`bool`
+        ``True`` to compute the smallest angle between the two vectors.
```
oesteban, line 36: `smallest` feels like a misnomer - This is just ensuring the maximum angle is 90 deg. by flipping the direction of vectors with larger angles, correct?

jhlegarreta: Agreed; have not been sharp enough to come up with a better name following the description in the paper. Would be grateful if you could come up a suggestion.

src/eddymotion/model/gradient_utils.py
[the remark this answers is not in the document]
jhlegarreta: All bvecs are known beforehand, and thus all angles can be computed beforehand, so it should be a matter of indexing the appropriate index and getting all corresponding pairwise values, right? Maybe I'm missing something, but under the loo convention, we would leave an index out, so we would be addressing the bvecs by their indices, isn't it? See this related https://github.com/nipreps/eddymotion/pull/176#discussion_r1584423393. Thus, a function could be called that takes the bvec index that has been left out, the angle matrix and picks the right angle values, isn't it?

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
...
+    Returns
+    -------
+    :obj:`float`
+        Angle between vectors.
+    """
+
+    dot_product = np.dot(v1, v2)
+    norm_v1 = np.linalg.norm(v1)
+    norm_v2 = np.linalg.norm(v2)
+    cosine_angle = dot_product / (norm_v1 * norm_v2)
```
oesteban, line 69: Perhaps it is clearer if we use the operator and ensure unit vectors first:

```suggestion
    cosine_angle = (v1 / np.linalg.norm(v1)) @ (v2 / np.linalg.norm(v2))
```

src/eddymotion/math/utils.py
oesteban:
```suggestion
    closest_polarity : :obj:`bool`
        If ``True``, the polarity of ``v2`` is reversed if the angle w.r.t. ``v1`` is smaller.
```

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
     except np.linalg.LinAlgError:
         # Matrix is not positive definite
         return False
+
+
+def compute_angle(v1, v2, smallest=False):
```
oesteban, line 48:
```suggestion
def compute_angle(v1, v2, closest_polarity=False):
```

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
...
+    """Compute the angle between two vectors.
+
+    Parameters
+    ----------
+    v1 : :obj:`ndarray`
```
oesteban, line 53:
```suggestion
    v1 : :obj:`~numpy.ndarray`
```

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
...
+        First vector.
+    v2 : :obj:`ndarray`
```
oesteban, line 55:
```suggestion
    v2 : :obj:`~numpy.ndarray`
```

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
...
+    :obj:`float`
+        Angle between vectors.
+    """
+
+    dot_product = np.dot(v1, v2)
+    norm_v1 = np.linalg.norm(v1)
+    norm_v2 = np.linalg.norm(v2)
+    cosine_angle = dot_product / (norm_v1 * norm_v2)
+    if smallest:
+        cosine_angle = np.abs(cosine_angle)
```
oesteban, line 71: Perhaps all the massaging should happen together?

```suggestion
    cosine_angle = np.clip(
        np.abs(cosine_angle) if closest_polarity else cosine_angle,
        -1.0,
        1.0,
    )
```

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
...
+    # Clip values to handle numerical errors
+    return np.arccos(np.clip(cosine_angle, -1.0, 1.0))
```
oesteban, line 73:
```suggestion
    return np.arccos(cosine_angle)
```

src/eddymotion/model/gradient_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        raise ValueError(f"bvecs must be of shape (3, N). Found: {bvecs.shape}")
+
+    num_vectors = bvecs.shape[1]
+    angles = np.zeros((num_vectors, num_vectors))
+
+    # ToDo
+    # Note that the below also computes the angle of a given vector with itself:
+    # Should those be removed?
+    # ToDo
+    # Vectorize and refactor: bvecs.T @ bvecs to get the outer product, or use
```
oesteban, line 72: I agree with this:

- Vectors should be normalized before coming in, but nonetheless, it would not hurt to normalize here again (only once).
- We can move `compute_angle()` to be under tests and use it to validate this calculation if we obtain cosines matrix in one shot (bvecs.T @ bvecs) as you said here.

src/eddymotion/model/gradient_utils.py
```diff
@@ -0,0 +1,81 @@
...
+        Pairwise angles across diffusion gradient encoding directions.
+
+    References
+    ----------
+    .. [Andersson15] J. L. R. Andersson. et al., An integrated approach to
+       correction for off-resonance effects and subject movement in diffusion MR
+       imaging, NeuroImage 125 (2016) 1063–1078
+    """
+
+    if bvecs.shape[0] != 3:
```
oesteban, line 62: Do not assume this is a numpy array :)

```suggestion
    if np.shape(bvecs)[0] != 3:
```

src/eddymotion/model/gradient_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    # Should those be removed?
+    # ToDo
+    # Vectorize and refactor: bvecs.T @ bvecs to get the outer product, or use
+    # a separate method to compute the angle between two vectors (starting from
+    # np.dot)
+    for i in range(num_vectors):
+        for j in range(i + 1, num_vectors):
+            angle = compute_angle(bvecs[:, i], bvecs[:, j], smallest)
+            angles[i, j] = angle
+            angles[j, i] = angle  # Symmetric matrix
```
oesteban, line 79:
```suggestion
    # Ensure b-vectors are unit-norm
    bvecs = np.array(bvecs) / np.linalg.norm(bvecs, axis=0)
    cosines = np.clip(bvecs.T @ bvecs, -1.0, 1.0)
    if closest_polarity:
        cosines = np.abs(cosines)
    angles = np.arccos(cosines)
```

oesteban (commented):
Some suggestions...

src/eddymotion/math/utils.py
jhlegarreta: Done in 3fce5be.

src/eddymotion/math/utils.py
jhlegarreta: :100:. Thanks. Changed in 3fce5be.

src/eddymotion/math/utils.py
jhlegarreta: :+1: Done in 3fce5be.

src/eddymotion/math/utils.py
jhlegarreta: Done in 3fce5be.

src/eddymotion/math/utils.py
jhlegarreta: Changed in 3fce5be.

src/eddymotion/math/utils.py
jhlegarreta: Changed in 3fce5be.

src/eddymotion/model/gradient_utils.py
jhlegarreta: > Vectors should be normalized before coming in, but nonetheless, it would not hurt to normalize here again (only once).

Done in https://github.com/nipreps/eddymotion/pull/200/files?diff=unified&w=0#diff-37c2c29d72b17bb336c5c519649dc18ff1f8c1ca0ddebde172b27ec208385935R59.

> We can move compute_angle() to be under tests and use it to validate this calculation if we obtain cosines matrix in one shot (bvecs.T @ bvecs) as you said here.

So used the inner product with the `@` operator and done as suggested 3fce5be. @arokem Oscar wanted to have it now, so done.

Removed all comments, including those about the multi-shell case, etc. We'll see how to deal with that when the time comes.

src/eddymotion/model/gradient_utils.py
jhlegarreta: Done in 3fce5be.

src/eddymotion/model/gradient_utils.py
jhlegarreta: Changed in 3fce5be.

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
...
+    v2 : :obj:`~numpy.ndarray`
+        Second vector.
+    closest_polarity : :obj:`bool`
+        ``True`` to compute the smallest angle between the two vectors.
+
+    Returns
+    -------
+    :obj:`float`
+        Angle between vectors.
+    """
```
oesteban, line 64: Let's add a quick unit test:

```suggestion
    
    Examples
    --------
    >>> compute_angle(
    ...     np.array((1.0, 0.0, 0.0)),
    ...     np.array((-1.0, 0.0, 0.0)),
    ... )  # doctest: +ELLIPSIS
    3.1415...
    >>> compute_angle(
    ...     np.array((1.0, 0.0, 0.0)),
    ...     np.array((-1.0, 0.0, 0.0)),
    ...     closest_polarity=True,
    ... )
    0.0
    
    """
```

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
...
+        First vector.
+    v2 : :obj:`~numpy.ndarray`
+        Second vector.
+    closest_polarity : :obj:`bool`
+        ``True`` to compute the smallest angle between the two vectors.
+
+    Returns
+    -------
+    :obj:`float`
+        Angle between vectors.
```
oesteban, line 63:
```suggestion
        The angle between the two vectors in radians.
```

src/eddymotion/math/utils.py
```diff
@@ -43,3 +43,31 @@ def is_positive_definite(matrix):
...
+    """Compute the angle between two vectors.
+
+    Parameters
+    ----------
+    v1 : :obj:`~numpy.ndarray`
+        First vector.
+    v2 : :obj:`~numpy.ndarray`
+        Second vector.
+    closest_polarity : :obj:`bool`
+        ``True`` to compute the smallest angle between the two vectors.
```
oesteban, line 58:
```suggestion
        ``True`` to reverse v2 if it points toward the opposite half-sphere.

File: src/eddymotion/model/gradient_utils.py [hunk 19]
```diff
@@ -0,0 +1,61 @@
...
+    bvecs : :obj:`~numpy.ndarray`
+        Diffusion gradient encoding directions in FSL format.
+    closest_polarity : :obj:`bool`
+        ``True`` to compute the smallest angle between each direction pair.
+
+    Returns
+    -------
+    :obj:`~numpy.ndarray`
+        Pairwise angles across diffusion gradient encoding directions.
+
```
oesteban, line 47:
```suggestion

    Examples
    --------
    >>> compute_pairwise_angles(
    ...     ((1.0, -1.0), (0.0, 0.0), (0.0, 0.0)),
    ...     False,
    ... )[0, 1]  # doctest: +ELLIPSIS
    3.1415...
    >>> compute_pairwise_angles(
    ...     ((1.0, -1.0), (0.0, 0.0), (0.0, 0.0)),
    ...     True,
    ... )[0, 1]
    0.0

```

oesteban (approved):
Suggesting some doctests, otherwise looks good!

src/eddymotion/math/utils.py
jhlegarreta: Added in d69bafd.

src/eddymotion/model/gradient_utils.py
jhlegarreta: Added in d69bafd.

src/eddymotion/math/utils.py
oesteban: It's okay as long as it is clear :)

However, if the angle surpasses 90 degrees, it means that, taking v1 as reference (and pole of the hemi-sphere), v2 is not in the hemi-sphere. Therefore, you reverse v2 and then the angle will be 45 deg. Without loss of generality, of course you can take v2 and set it as the pole, and everything works the same.

oesteban: i.e., you don't define the hemispheres as positive/negative, but rather around the reference vector:

Boston is in the northern hemisphere, as is Singapore. However, if you define the hemispheres around the first vector (Boston), then Singapore is in the opposite hemisphere because their angle is more than 90 deg.

jhlegarreta: I do not see how a hemi-sphere is defined then: v1 and v2 can be in the same "north" hemi-sphere and have a 135° angle; if we reverse v2 and take 45°, they will not be in the "north" hemi-sphere, but will be in another hemi-sphere, e.g. the "east" one... So the hemi-sphere thing here is confusing to me. I may still be mistaken.

oesteban: Sure -- let's keep it as you wrote it :)

PR #201, opened 2024-06-02
ENH: Add squared exponential covariance kernel

jhlegarreta:
Add squared exponential covariance kernel.


jhlegarreta:
The remaining comments in https://github.com/nipreps/eddymotion/pull/188#issuecomment-2118215168 and https://github.com/jhlegarreta/eddymotion/pull/1 will be addressed in this PR.

oesteban:
I've added the valuable text in the docstrings into #216 so they are not lost. When we implement multi-shell, we can come back to this one for inspiration on the missing bits.

Closing in favor of the omnibus PR (#216). Please reopen if that doesn't feel okay.

PR #202, opened 2024-06-04
ENH: Add gaussian process DWI signal representation notebook

jhlegarreta:
Add gaussian process DWI signal representation notebook.


jhlegarreta:
Depends on #188. @esavary If you did changes to the notebook, maybe you can do a PR to this branch with such changes, and I would merge it right away.

With the default kernels, I get this error plot, so the default kernels are not a good choice for the DWI signal:

gaussian_process_default_kernel

Not sure if the signal not being normalized plays a role here.

esavary:
I made a pull request against your branch with some suggestions. Essentially, it involves not fitting one voxel at a time. This ensures that the same hyperparameters are used for the kernel across all voxels. It also improves their estimation when done with more voxels (this is done inside the fit method). These changes result in a slight improvement. However, even with these changes, the optimization is still failing for the parameter k2__noise_level (from the WhiteNoise kernel). It hits the bounds, leading to a significant error, as shown in this plot (I used the default hyperparameters optimization from the fit method of sklearn not the custom one I coded):

results_small_cube

Additionally, I tried fitting a larger cube, but it seems to be similar as with the SphericalKernel. See the image below showing the prediction:

estimationwithekernel

An additional comment I have is that we might consider using the angles between the directions instead of bvecs directly, as suggested in Andersson (2015). However, I was unsure if that was only relevant when using the Spherical or Exponential kernel, so I did not include that change.

esavary:
> Depends on #188. @esavary If you did changes to the notebook, maybe you can do a PR to this branch with such changes, and I would merge it right away.
> 
> With the default kernels, I get this error plot, so the default kernels are not a good choice for the DWI signal:
> 
> gaussian_process_default_kernel
> 
> Not sure if the signal not being normalized plays a role here.



> Depends on #188. @esavary If you did changes to the notebook, maybe you can do a PR to this branch with such changes, and I would merge it right away.
> 
> With the default kernels, I get this error plot, so the default kernels are not a good choice for the DWI signal:
> 
> gaussian_process_default_kernel
> 
> Not sure if the signal not being normalized plays a role here.

Actually, I've just noticed something in your plot,  I think the ground truth signal should not be zero (but not sure if it causes the problem). Could it be that the data you used for the test is empty except for the first gradient direction? ( I used this data for my plot: https://osf.io/download/6at98/)

jhlegarreta:
Merged your PR into the branch.

Maybe we can now uncomment the orthoview calls and provide with the same values you used to generate the figures you pasted.

> An additional comment I have is that we might consider using the angles between the directions instead of bvecs directly, as suggested in Andersson (2015). However, I was unsure if that was only relevant when using the Spherical or Exponential kernel, so I did not include that change.

In my understanding we want to predict on bvecs. We want to represent a given DWI volume using a GP, e.g. as if we were using another basis (e.g. SH) to represent it and be able to recreate it within some tolerance. The ISMRM paper/poster figure is helpful in this.

> Actually, I've just noticed something in your plot, I think the ground truth signal should not be zero (but not sure if it causes the problem). Could it be that the data you used for the test is empty except for the first gradient direction? ( I used this data for my plot: https://osf.io/download/6at98/)

Maybe the volume I chose included some background voxels.

jhlegarreta:
Maybe we should normalize the DWI signal (divide each voxel signal by the B0 signal in that location) prior to fitting the data. Not sure if this would have a large influence on the results, but maybe worthwhile trying.

jhlegarreta:
Added a new notebook with a simulated signal using a multi-tensor model.

@arokem In an effort to get something like fig 4 in Andersson 2015 NIMG, I am getting the below figure for the signal I am simulating in `docs/notebooks/dwi_simulated_gp.ipynb`. 

dwi_multi-tensor_signal_90dirs

I would expects the diffusion signals to roughly describe the two crossing tensors, but what I get is a blob. Tensors are at 90 degrees, and the fODF I get from them shows so:

dwi_multi-tensor_signal_90dirs_fodf

What am I missing? Thanks.

arokem:
Sorry for the slowness. One quick thought is that Andersson et al. used a b=3,000 in their simulations and I see that you are using b=1,000. Another is that I suspect that the Andersson figures were generated with a transformation to diffusion distance, rather than in the original signal domain. Indeed, a crossing fiber signal would look like the linear combination of two "donuts" rather than the linear combination of two "cigars", because the _signal_ looks like a donut (see e.g., figure 1 in [my 2014 paper](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0123272)), and when you add these donuts together, especially in low b-values, you get something that more closely resembles a "blob". To get "cigars" you need to do something like this: https://github.com/arokem/osmosis/blob/72d848ddc172ce6a72baba4c0d4cf3675db1a381/osmosis/tensor.py#L233-L248

jhlegarreta:
Removed the signal plot (https://github.com/nipreps/eddymotion/pull/202#issuecomment-2170423547) for now so that it does not distract us as we progress with the kernels, etc. Will add it in a separate PR on top of this branch.

oesteban:
@jhlegarreta can you commit with the cells' output so that we can use github to visualize the results? It's tough to look at the diff, but we workaround the problem of the HCP data.

jhlegarreta:
> @jhlegarreta can you commit with the cells' output so that we can use github to visualize the results? It's tough to look at the diff, but we workaround the problem of the HCP data.

Brought back the cell in the last push force. Note that in PR #217 I had tried with the approach Ariel suggested but did not get the expected shape.

docs/notebooks/dwi_simulated_gp.ipynb
```diff
@@ -0,0 +1,685 @@
...
+     "end_time": "2024-07-12T01:48:51.227899Z",
+     "start_time": "2024-07-12T01:48:51.205608Z"
+    }
+   },
+   "cell_type": "code",
+   "source": [
+    "# Create an interactive plot\n",
+    "%matplotlib notebook\n",
+    "\n",
+    "fig = plot_dwi_scatter(signal_repr)\n",
```
jhlegarreta, line 302: Shows two disks, so not quite right yet.

signal_scatter_2

signal_scatter_1

docs/notebooks/dwi_simulated_gp.ipynb
```diff
@@ -0,0 +1,685 @@
...
+    }
+   ],
+   "execution_count": 9
+  },
+  {
+   "metadata": {},
+   "cell_type": "code",
+   "source": [
+    "# ToDo\n",
+    "# Plot as surface"
```
jhlegarreta, line 337: Have not succeeded in getting this right.

docs/notebooks/dwi_simulated_gp.ipynb
```diff
@@ -0,0 +1,685 @@
...
+    "    s_hat_mean + 1.96 * s_hat_stddev,\n",
+    "    alpha=0.5,\n",
+    "    color=\"orange\",\n",
+    "    label=r\"95% confidence interval\",\n",
+    ")\n",
+    "plt.scatter(x, s, c=\"b\", label=\"ground truth\")\n",
+    "ax.set_xlabel(\"bvec indices\")\n",
+    "ax.set_ylabel(\"signal\")\n",
+    "ax.legend()\n",
+    "plt.title(\"Gaussian process regression on dataset\")\n",
```
jhlegarreta, line 489: The plot looks perfect because I am predicting on the very same data that I used for fitting.

jhlegarreta:
A few notes:
- Have not investigated the optimization convergence error shown on the notebok (`ABNORMAL_TERMINATION_IN_LNSRCH.`): when training and predicting on all bvecs (except b0) the warning was being shown, but not the error.
- Have the surface plots of the diffusion signal (which could be used for the prediction) in a separate piece of code. When we take the time to discuss about how confident we are about the GP (e.g. the differences wrt the previous implementation, etc.), and other issues (example organisation, data and code to generate the empirical covariance plot) we can bring that part.

oesteban:
Is there anything I can do here before the TechMon meeting?

jhlegarreta:
@oesteban I apologise for not having worked on this in the last few weeks; I expect to be able to have another look around the tech meeting date. Sorry.

The single point prediction looks to be working, but I am not sure about the source of the error message mentioned in https://github.com/nipreps/eddymotion/pull/202#issuecomment-2282914257, so I need to investigate that. I would like to pick other random b-vecs to see if the prediction is OK or whether the convergence message persists.

Additionally, I would like to investigate what has changed wrt the implementation that was not working. I need to prepare 
 multi-fiber settings, and add the predicted point visualization part. So all this is on my roof; if you feel like, you can run the notebook and see if my impression that it is working seems OK (despite the convergence message).

Probably we can talk about all these experiments/design (esp. when we will add experiments with acquired data) in our eventual meeting.

jhlegarreta:
As things stand now, I would split the PR into 2 PRs: one with the notebook that uses a simulated fiber and another one the one that uses real data. The first one is self-contained and finished IMO: I would not complicate more the notebook.

oesteban:
> As things stand now, I would split the PR into 2 PRs: one with the notebook that uses a simulated fiber and another one the one that uses real data. The first one is self-contained and finished IMO: I would not complicate more the notebook.

I agree with this take. Should we close this one and go ahead with two fresh PRs? (feel free to reopen if I rushed closing).

PR #203, opened 2024-06-05
ENH: add spherical model kernel

esavary:
- Integrate custom kernel based on spherical covariance function.
- Add tests for the kernel.

### DIFF
--- a/src/eddymotion/model/kernels.py
+++ b/src/eddymotion/model/kernels.py
@@ -30,14 +30,14 @@ class SphericalCovarianceKernel(Kernel):
 
     Parameters
     ----------
-    lambda_ : float, default=1.0
+    lambda_s : float, default=1.0
         Scale parameter for the covariance function.
     a : float, default=1.0
         Distance parameter where the covariance function goes to zero.
     sigma_sq : float, default=1.0
         Noise variance term.
-    lambda_bounds : pair of floats >= 0 or "fixed", default=(1e-5, 1e4)
-        The lower and upper bound on lambda_.
+    lambda_s_bounds : pair of floats >= 0 or "fixed", default=(1e-5, 1e4)
+        The lower and upper bound on lambda_s.
     a_bounds : pair of floats >= 0 or "fixed", default=(1e-5, np.pi)
         The lower and upper bound on a.
     sigma_sq_bounds : pair of floats >= 0 or "fixed", default=(1e-5, 1e4).
@@ -45,23 +45,23 @@ class SphericalCovarianceKernel(Kernel):
 
     def __init__(
         self,
-        lambda_=2.0,
+        lambda_s=2.0,
         a=0.1,
         sigma_sq=1.0,
-        lambda_bounds=(1e-5, 1e4),
+        lambda_s_bounds=(1e-5, 1e4),
         a_bounds=(1e-5, np.pi),
         sigma_sq_bounds=(1e-5, 1e4),
     ):
-        self.lambda_ = lambda_
+        self.lambda_s = lambda_s
         self.a = a
         self.sigma_sq = sigma_sq
-        self.lambda_bounds = lambda_bounds
+        self.lambda_s_bounds = lambda_s_bounds
         self.a_bounds = a_bounds
         self.sigma_sq_bounds = sigma_sq_bounds
 
     @property
-    def hyperparameter_lambda(self):
-        return Hyperparameter("lambda_", "numeric", self.lambda_bounds)
+    def hyperparameter_lambda_s(self):
+        return Hyperparameter("lambda_s", "numeric", self.lambda_s_bounds)
 
     @property
     def hyperparameter_a(self):
@@ -104,12 +104,12 @@ class SphericalCovarianceKernel(Kernel):
             dists_deriv[mask] = (3 * theta[mask] / self.a**2) - (
                 1.5 * (theta[mask] / self.a) ** 2
             ) / self.a
-            K_gradient[:, :, 0] = K / self.lambda_
-            K_gradient[:, :, 1] = self.lambda_ * dists_deriv
+            K_gradient[:, :, 0] = K / self.lambda_s
+            K_gradient[:, :, 1] = self.lambda_s * dists_deriv
             K_gradient[:, :, 2] = np.eye(len(theta))
-            return self.lambda_ * K + self.sigma_sq * np.eye(len(theta)), K_gradient
+            return self.lambda_s * K + self.sigma_sq * np.eye(len(theta)), K_gradient
         else:
-            return self.lambda_ * K + self.sigma_sq * np.eye(len(theta))
+            return self.lambda_s * K + self.sigma_sq * np.eye(len(theta))
 
     def diag(self, X):
         """
@@ -125,7 +125,7 @@ class SphericalCovarianceKernel(Kernel):
         array-like of shape (n_samples,)
             Diagonal of the kernel matrix.
         """
-        return np.full(X.shape[0], self.lambda_ + self.sigma_sq)
+        return np.full(X.shape[0], self.lambda_s + self.sigma_sq)
 
     def is_stationary(self):
         """
@@ -152,7 +152,7 @@ class SphericalCovarianceKernel(Kernel):
         params : dict
             Parameter names mapped to their values.
         """
-        return {"lambda_": self.lambda_, "a": self.a, "sigma_sq": self.sigma_sq}
+        return {"lambda_s": self.lambda_s, "a": self.a, "sigma_sq": self.sigma_sq}
 
     def set_params(self, **params):
         """
@@ -168,7 +168,7 @@ class SphericalCovarianceKernel(Kernel):
         self : object
             Returns self.
         """
-        self.lambda_ = params.get("lambda_", self.lambda_)
+        self.lambda_s = params.get("lambda_s", self.lambda_s)
         self.a = params.get("a", self.a)
         self.sigma_sq = params.get("sigma_sq", self.sigma_sq)
         return self
--- a/src/eddymotion/model/kernels.py
+++ b/src/eddymotion/model/kernels.py
@@ -101,9 +101,7 @@ class SphericalCovarianceKernel(Kernel):
         """
         if theta_prime is not None:
             theta = theta_prime
-        theta = np.atleast_2d(theta)
-        K = np.where(theta <= self.a, 1 - 1.5 * (theta / self.a) + 0.5 * (theta / self.a)**3, 0)
-        K = self.lambda_ * K + self.sigma_sq * np.eye(len(theta))
+        K = np.where(theta <= self.a, 1 - 3 * (theta / self.a) ** 2 + 2 * (theta / self.a) ** 3, 0)
 
         if eval_gradient:
             K_gradient = np.zeros((theta.shape[0], theta.shape[1], 3))
@@ -113,9 +111,9 @@ class SphericalCovarianceKernel(Kernel):
             K_gradient[:, :, 0] = K / self.lambda_
             K_gradient[:, :, 1] = self.lambda_ * dists_deriv
             K_gradient[:, :, 2] = np.eye(len(theta))
-            return K, K_gradient
+            return self.lambda_ * K + self.sigma_sq * np.eye(len(theta)), K_gradient
         else:
-            return K
+            return self.lambda_ * K + self.sigma_sq * np.eye(len(theta))
 
     def diag(self, X):
         """
--- a/src/eddymotion/model/kernels.py
+++ b/src/eddymotion/model/kernels.py
@@ -56,8 +56,15 @@ class SphericalCovarianceKernel(Kernel):
         The lower and upper bound on 'sigma_sq'.
     """
 
-    def __init__(self, lambda_=2.0, a=0.1, sigma_sq=1.0,
-                 lambda_bounds=(1e-5, 1e4), a_bounds=(1e-5, np.pi), sigma_sq_bounds=(1e-5, 1e4)):
+    def __init__(
+        self,
+        lambda_=2.0,
+        a=0.1,
+        sigma_sq=1.0,
+        lambda_bounds=(1e-5, 1e4),
+        a_bounds=(1e-5, np.pi),
+        sigma_sq_bounds=(1e-5, 1e4),
+    ):
         self.lambda_ = lambda_
         self.a = a
         self.sigma_sq = sigma_sq
@@ -107,7 +114,9 @@ class SphericalCovarianceKernel(Kernel):
             K_gradient = np.zeros((theta.shape[0], theta.shape[1], 3))
             dists_deriv = np.zeros_like(theta)
             mask = theta <= self.a
-            dists_deriv[mask] = (3 * theta[mask] / self.a**2) - (1.5 * (theta[mask] / self.a)**2) / self.a
+            dists_deriv[mask] = (3 * theta[mask] / self.a**2) - (
+                1.5 * (theta[mask] / self.a) ** 2
+            ) / self.a
             K_gradient[:, :, 0] = K / self.lambda_
             K_gradient[:, :, 1] = self.lambda_ * dists_deriv
             K_gradient[:, :, 2] = np.eye(len(theta))
--- a/src/eddymotion/model/kernels.py
+++ b/src/eddymotion/model/kernels.py
@@ -1,0 +1,114 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2022 The NiPreps Developers <nipreps@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY kIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+import numpy as np
+from sklearn.gaussian_process.kernels import Kernel
+
+
+class SphericalCovarianceKernel(Kernel):
+    """
+    Custom kernel based on spherical covariance function.
+    Parameters
+    ----------
+    lambda_ : float, default=1.0
+        Scale parameter for the covariance function.
+    a : float, default=1.0
+        Distance parameter where the covariance function goes to zero.
+    sigma_sq : float, default=1.0
+        Noise variance term.
+    """
+    def __init__(self, lambda_=1.0, a=1.0, sigma_sq=1.0):
+        self.lambda_ = lambda_
+        self.a = a
+        self.sigma_sq = sigma_sq
+
+    def __call__(self, theta, theta_prime=None):
+        """
+        Compute the kernel matrix.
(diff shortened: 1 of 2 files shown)


src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,174 @@
...
+        Returns
+        -------
+        K : array-like of shape (n_samples, n_samples)
+            Kernel matrix.
+        K_gradient : array-like of shape (n_samples, n_samples, n_dims), optional
+            The gradient of the kernel matrix with respect to the log of the
+            hyperparameters. Only returned when `eval_gradient` is True.
+        """
+        if theta_prime is not None:
+            theta = theta_prime
```
oesteban, line 97: This doesn't seem very useful - can we remove this `theta_prime` altogether?

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Parameters
+        ----------
+        theta : array-like of shape (n_samples, n_samples)
+            Precomputed pairwise angles.
+        theta_prime : array-like of shape (n_samples, n_samples), default=None
+            Second input for the kernel function.
+        eval_gradient : bool, default=False
+            Determines whether the gradient with respect to the log of
+            the kernel hyperparameter is computed.
```
oesteban, line 86:
```suggestion
            Evaluate the gradient with respect to the log of
            the kernel hyperparameter.
```

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,174 @@
...
+            mask = theta <= self.a
+            dists_deriv[mask] = (3 * theta[mask] / self.a**2) - (
+                1.5 * (theta[mask] / self.a) ** 2
+            ) / self.a
+            K_gradient[:, :, 0] = K / self.lambda_s
+            K_gradient[:, :, 1] = self.lambda_s * dists_deriv
+            K_gradient[:, :, 2] = np.eye(len(theta))
+            return self.lambda_s * K + self.sigma_sq * np.eye(len(theta)), K_gradient
+        else:
+            return self.lambda_s * K + self.sigma_sq * np.eye(len(theta))
```
oesteban, line 112: The output of the function should be regular:

```suggestion
        K_gradient = None
        if eval_gradient:
            K_gradient = np.zeros((theta.shape[0], theta.shape[1], 3))
            dists_deriv = np.zeros_like(theta)
            mask = theta <= self.a
            dists_deriv[mask] = (3 * theta[mask] / self.a**2) - (
                1.5 * (theta[mask] / self.a) ** 2
            ) / self.a
            K_gradient[:, :, 0] = K / self.lambda_s
            K_gradient[:, :, 1] = self.lambda_s * dists_deriv
            K_gradient[:, :, 2] = np.eye(len(theta))

        return self.lambda_s * K + self.sigma_sq * np.eye(len(theta)), K_gradient
```

oesteban (commented):
I had a very quick pass and left a few suggestions.

src/eddymotion/model/kernels.py
esavary: This is useful when making predictions. The predict method of the Gaussian Regressor of sklearn calls the kernel with self.X_train in addition to the data for which we are making predictions (see [line 439](https://github.com/scikit-learn/scikit-learn/blob/5491dc695/sklearn/gaussian_process/_gpr.py#L439)). The assumption that theta equals theta_prime holds only if we feed our predict method directly with the matrix containing the angles between the directions of the training set and the direction we want to predict. If we decide to feed our predict method with something else, we will need to adjust this and compute the angles within the kernel.

oesteban: By looking at the interface: https://github.com/scikit-learn/scikit-learn/blob/5bc86b885c2d4c9afde3b41ebbee46dcf8ca4254/sklearn/gaussian_process/kernels.py#L400

And the documentation there (for the compound kernel)

```
        X : array-like of shape (n_samples_X, n_features) or list of object, \
            default=None
            Left argument of the returned kernel k(X, Y)

        Y : array-like of shape (n_samples_X, n_features) or list of object, \
            default=None
            Right argument of the returned kernel k(X, Y). If None, k(X, X)
            is evaluated instead.
```

By reassigning this, you are  basically enforcing k(Y, Y) which I don't think makes sense.

esavary: It makes sense when the kernel arguments are the matrices representing the differences between the bvec directions, not the bvec directions themselves. In this case, when making predictions, Y = θ(bvecs_test, bvecs_train). Therefore the kernel corresponds to k(x*, x) in equation 5 from Andersson 2015. However, I'm open to moving the computation of the angles inside the kernel and taking directly the bvec as arguments to make it more intuitive. Hence my comments [here](https://github.com/nipreps/eddymotion/pull/200#discussion_r1620664936) and [here](https://github.com/jhlegarreta/eddymotion/pull/2#discussion_r1613178501)

oesteban: > It makes sense when the kernel arguments are the matrices representing the differences between the bvec directions, not the bvec directions themselves. In this case, when making predictions, Y = θ(bvecs_test, bvecs_train). Therefore the kernel corresponds to k(x*, x) in equation 5 from [Andersson 2015](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4627362/).

Scikit learn is going to do k(x*, x) all by itself if Y=None -- the question is how you call kernel().

Now, if you want X to be only train samples, then something external to this PR will do just that (I would assume within the Gaussian Process PR).

I would be very surprised if this is the right place to do this.

oesteban: Instead of silently replacing the X input, we should do something explicit like this:

```suggestion
        if Y is not None:
            raise RuntimeError('Y should not be set')
```

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,174 @@
...
+        -------
+        K : array-like of shape (n_samples, n_samples)
+            Kernel matrix.
+        K_gradient : array-like of shape (n_samples, n_samples, n_dims), optional
+            The gradient of the kernel matrix with respect to the log of the
+            hyperparameters. Only returned when `eval_gradient` is True.
+        """
+        if theta_prime is not None:
+            theta = theta_prime
+        K = np.where(theta <= self.a, 1 - 3 * (theta / self.a) ** 2 + 2 * (theta / self.a) ** 3, 0)
```
oesteban, line 98: Please update the argument throughout after the function definition was changed:

```suggestion
        K = np.where(X <= self.a, 1 - 3 * (X / self.a) ** 2 + 2 * (X / self.a) ** 3, 0)
```

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,174 @@
...
+
+        K_gradient = None
+        if eval_gradient:
+            K_gradient = np.zeros((theta.shape[0], theta.shape[1], 3))
```
oesteban, line 102: This feels odd. X should be (n_samples, n_features), and for DWI this is (N, 3) with N gradient directions. I think the gradient differences indeed must be calculated inside the kernel.


```suggestion
            K_gradient = np.zeros((X.shape[0], X.shape[1], 3))
```

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+
+    def __init__(
+        self,
+        lambda_s=2.0,
+        a=0.1,
+        sigma_sq=1.0,
+        lambda_s_bounds=(1e-5, 1e4),
+        a_bounds=(1e-5, np.pi),
+        sigma_sq_bounds=(1e-5, 1e4),
```
jhlegarreta, line 53: Maybe I did not revise the papers or the references thoroughly, but would be interested in knowing where the defaults come from.

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,174 @@
...
+from sklearn.gaussian_process.kernels import Hyperparameter, Kernel
+
+
+class SphericalCovarianceKernel(Kernel):
+    """
+    Custom kernel based on spherical covariance function.
+
+    Parameters
+    ----------
+    lambda_s : float, default=1.0
```
jhlegarreta, line 33: Would remove the defaults from docstrings. They are documented in the method signature.

We have used `:obj:float` (generally `:obj:name`) to document all types.

test/test_model_kernels.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Compute the kernel matrix using the kernel instance
+    K = kernel(theta)
+
+    # Assert the kernel matrix is as expected
+    np.testing.assert_array_almost_equal(K, expected_K, decimal=6)
+
+    # Check if the kernel matrix is positive definite
+    eigenvalues = np.linalg.eigvals(K)
+    assert np.all(eigenvalues > 0), "Kernel matrix is not positive definite"
```
jhlegarreta, line 92: Would call directly https://github.com/nipreps/eddymotion/blob/main/src/eddymotion/math/utils.py#L26.

test/test_model_kernels.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Compute the diagonal using the kernel instance
+    diag = kernel.diag(X)
+
+    # Assert the diagonal is as expected
+    np.testing.assert_array_almost_equal(diag, expected_diag, decimal=6)
+
+
+if __name__ == "__main__":
+    pytest.main()
```
jhlegarreta, line 113: Not needed https://github.com/nipreps/eddymotion/pull/203/files#diff-c6a0d4f84e7e55e221b504f1c82912f0408b0321dca13e6f15d203443e0cce50R111-R113.

src/eddymotion/model/kernels.py
jhlegarreta: I may be missing something, but I do believe this https://github.com/nipreps/eddymotion/pull/203#discussion_r1631931390 is right.

`*x*` (and `*x'*`) are our bvecs, so I do not see where `Y` comes into play in the kernel. If we still disagree, maybe we can talk about it over a call and draw on a screen why we believe one thing of the other.

jhlegarreta (commented):
Thanks for doing this @esavary.

Once the two outstanding comments are solved, this is ready to go IMO.

Further improvements (including my own comments) can be addressed in separate PRs.

src/eddymotion/model/kernels.py
esavary: >Now, if you want X to be only train samples, then something external to this PR will do just that (I would assume within the Gaussian Process PR) 
Yes, I assumed we would call the fit method of GaussianRegressor with the training set and the predict method with the test set using the remaining bvecs. When we do this, inside the [predict method](https://github.com/scikit-learn/scikit-learn/blob/5491dc695/sklearn/gaussian_process/_gpr.py#L439) of the GaussianRegressor, the kernel is called with two arguments: self.x_train and Y. If we directly input the angles difference, not the bvecs, Y is θ(bvecs_test, bvecs_train) (the angle difference between the bvec we want to predict and the bvecs from the training set. 

However, I strongly agree that it would be better to compute the angles inside the kernel. Additionally, if we input the angles difference directly into the predict method of sklearn, we will get a prediction for each angle. What we want, however, is a single prediction for each bvec. 

@jhlegarreta Sorry, it's a bit difficult to explain the problem with the dimensions in a github issue. I think organizing a meeting with a virtual whiteboard is a very good idea.

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,174 @@
...
+    def __call__(self, theta, theta_prime=None, eval_gradient=False):
+        """
+        Compute the kernel matrix.
+
+        Parameters
+        ----------
+        theta : array-like of shape (n_samples, n_samples)
+            Precomputed pairwise angles.
+        theta_prime : array-like of shape (n_samples, n_samples), default=None
+            Second input for the kernel function.
```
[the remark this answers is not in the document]
esavary: Let's first discuss with @jhlegarreta  about moving the angle computation inside. From this, we will know precisely what Y is

src/eddymotion/model/kernels.py
esavary: I made an initial rough guess for the values using Fig. 2 from Andersson 2015, but I only found the value for a in the paper. I think σ could potentially be estimated based on the noise level, but I'm not sure of the best approach (perhaps only in grey matter voxels). Do you have any input on this? For the bounds, I used a half sphere for a and set the upper bound of λ and σ to one order of magnitude lower than the usual defaults in sklearn, but we might be able to determine smarter bounds.

test/test_model_kernels.py
esavary: Done in https://github.com/nipreps/eddymotion/pull/203/commits/ea6baecd70471850bd208200fe121247fac4591a

test/test_model_kernels.py
esavary: Done in https://github.com/nipreps/eddymotion/pull/203/commits/ea6baecd70471850bd208200fe121247fac4591a

src/eddymotion/model/kernels.py
esavary: I reverted this change in [9d0a7c6](https://github.com/nipreps/eddymotion/pull/203/commits/9d0a7c67dfadd26d38b4af8a62e3a1eeb25a609e). Other kernels from the sklearn library don't produce regular output( see[ kernels](https://github.com/scikit-learn/scikit-learn/blob/2621573e6/sklearn/gaussian_process/kernels.py) ), and returning the gradient at any time causes errors when the kernel is called with eval_gradients=False.

src/eddymotion/model/kernels.py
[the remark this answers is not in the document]
jhlegarreta: > about moving the angle computation inside

Talked to Óscar yesterday about this, and we concluded that my initial thought of providing the kernel with the angles is probably not right. So I take the blame for having pushed things into that direction.

Óscar mentioned that this would allow the GP to do the below computation appropriately when providing the query bvec in the prediction:
https://github.com/scikit-learn/scikit-learn/blob/972e17fe1aa12d481b120ad4a3dc076bae736931/sklearn/gaussian_process/_gpr.py#L439

So the kernel should receive the bvecs, and make the computation of the angles just before these lines:
https://github.com/nipreps/eddymotion/pull/203/files#diff-7401613f4c745123dea57ac1f74163b4e14f6ed2e0c82fa2e466835139743fdbR99-R100 and https://github.com/nipreps/eddymotion/pull/203/files#diff-7401613f4c745123dea57ac1f74163b4e14f6ed2e0c82fa2e466835139743fdbR103-R105

e.g.
```
X = compute_pairwise_angles(bvecs.T, closest_polarity=True)
K = self.lambda_s * np.where(
    X <= self.a, 1 - 3 * (X / self.a) ** 2 + 2 * (X / self.a) ** 3, 0
) + self.sigma_sq * np.eye(len(X))
```

Doing the above in an example with a simulated signal (using https://github.com/nipreps/eddymotion/blob/a8742d955d9a43676d54ef194737a42db6f74015/docs/notebooks/dwi_simulated_gp.ipynb) triggers a dimensionality error on `K_gradient` by the underlying call 
https://github.com/scikit-learn/scikit-learn/blob/972e17fe1aa12d481b120ad4a3dc076bae736931/sklearn/gaussian_process/_gpr.py#L297

So the computation of `K_gradient` will need to be adapted now. 

The dimensionality issues that you talked about in other messages is probably related to this. Fixing the above will hopefully will put us in the right track.

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,186 @@
...
+        Returns
+        -------
+        K : :obj:`array-like` of shape (n_samples, n_samples)
+            Kernel matrix.
+        K_gradient : :obj:`array-like` of shape (n_samples, n_samples, n_dims), optional
+            The gradient of the kernel matrix with respect to the log of the
+            hyperparameters. Only returned when `eval_gradient` is True.
+        """
+        if Y is not None:
+            X = Y
```
jhlegarreta, line 98: So I see that in https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.kernels.Kernel.html they do the `X=Y` assignment if `Y is None` (note the absence of the `not`).

What if we avoid doing the `X=Y` assignment to get past this situation? My thought is linked to 
This is related to https://github.com/nipreps/eddymotion/pull/203/files#r1641440923, and I now completely see your point in https://github.com/nipreps/eddymotion/pull/200#discussion_r1615824997 @esavary:

```
if Y is None:
  theta = compute_pairwise_angles(X, X)
else:
  theta = compute_pairwise_angles(X, Y)

 K = self.lambda_s * np.where(
                theta <= self.a, 1 - 3 * (theta / self.a) ** 2 + 2 * (theta / self.a) ** 3, 0
            )
```

(assuming here `X` and potentially `Y` are a set of gradient vectors provided by the caller.)

(Have not checked whether we would still need to multiply by `np.eye(len(theta)` if `Y is None`)

The `eval_gradient` block may need to be adjusted concerning the dimensions now that `X` is assumed to be `bvecs`.

So `compute_pairwise_angles` would need to be modified to accept two input parameters. I apologize for not having realized about the downstream needs and implications of these two aspects. Really sorry @esavary.

esavary: > So `compute_pairwise_angles` would need to be modified to accept two input parameters. I apologize for not having realized about the downstream needs and implications of these two aspects. Really sorry @esavary.
No worries, it also makes a lot of sense to compute the angles beforehand since we have all the information at the beginning. It would just have required us to recode the fit method. I'll suggest a generalization of the angle computation in a separate PR and then use this to update the kernel to take the bvec directly. This will also solve the problem of X=Y because we can then perform different computations when Y is given.

jhlegarreta: PR #212 to make `compute_pairwise_angles` accept two parameters.

esavary:
I turned the PR into a draft for the moment because it depends on PR #212 . I integrated the angle computation into the kernel using the new function from that PR.

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,190 @@
...
+        -------
+        K : :obj:`~numpy.ndarray` of shape (n_samples, n_samples)
+            Kernel matrix.
+        K_gradient : :obj:`~numpy.ndarray` of shape (n_samples, n_samples, n_dims), optional
+            The gradient of the kernel matrix with respect to the log of the
+            hyperparameters. Only returned when `eval_gradient` is True.
+        """
+        if Y is not None:
+            angles = compute_pairwise_angles(X.T, Y.T, closest_polarity=True)
+            K = self.lambda_s * np.where(
```
jhlegarreta, line 101: We probably want to call this
https://github.com/nipreps/eddymotion/pull/213/files#diff-ba53cdfdd502439a31a24c148f9be58dab5fb56d09b25547557933b3b259e46fR57

Making that a function will allow for reusing it in the multi-shell case. Also, will allow us to ensure that the behavior is the expected one prior to the kernel computations (e.g. together with PR #214).

src/eddymotion/model/kernels.py
```diff
@@ -0,0 +1,188 @@
...
+            hyperparameters. Only returned when `eval_gradient` is True.
+        """
+        if Y is not None:
+            angles = compute_pairwise_angles(X.T, Y.T, closest_polarity=True)
+            cov_matrix = compute_spherical_covariance(angles, self.a)
+            K = self.lambda_s * cov_matrix
+        else:
+            angles = compute_pairwise_angles(X.T, X.T, closest_polarity=True)
+            cov_matrix = compute_spherical_covariance(angles, self.a)
+            K = self.lambda_s * cov_matrix + self.sigma_sq * np.eye(len(angles))
```
jhlegarreta, line 106: This can probably be simplified since the pairwise angle computation now accepts the second argument to be `None`. Also, looking at $\mathbf{K}_{y} = \mathbf{K} + \sigma^2 \mathbf{I}$ just below eq. 12 in Andresson15 maybe `self.sigma_sq * np.eye(len(angles))` needs to be added in either case, and thus we would not need the `if/else` statement.

oesteban:
Superseded by #227 and other efforts. Good to have around for the interesting discussion above.

PR #204, opened 2024-06-06
ENH: Custom Optimizers for Hyperparameter Optimization

esavary:
- Introduces custom optimization methods for hyperparameter tuning in Gaussian Process models, inspired by Andersson et al. (2015):
  -  Negative Log-Likelihood Optimization
  -  Leave-One-Out Cross-Validation (LOO-CV)
- Includes test with standard kernels from scikit-learn.

### DIFF
--- a/src/eddymotion/model/optimizers.py
+++ b/src/eddymotion/model/optimizers.py
@@ -86,7 +86,7 @@ def total_negative_log_likelihood(beta, y_all, X, kernel, reg_param=1e-6):
         Total negative log marginal likelihood.
     """
     total_log_likelihood = 0
-    for y in y_all.T:  # Iterate over voxels
+    for y in y_all.T:  #transposed to have shape (n_voxels, n_samples) and iterate over voxels
         total_log_likelihood += negative_log_likelihood(beta, y, X, kernel, reg_param)
     return total_log_likelihood
 
--- a/src/eddymotion/model/optimizers.py
+++ b/src/eddymotion/model/optimizers.py
@@ -58,7 +58,9 @@ def negative_log_likelihood(beta, y, X, kernel, reg_param=1e-6):
         print("Non-positive definite kernel matrix")
         return 1e10  # Penalize non-positive definite kernel
 
-    log_likelihood = -0.5 * (np.dot(y.T, np.linalg.solve(K, y)) + np.linalg.slogdet(K)[1] + len(y) * np.log(2 * np.pi))
+    log_likelihood = -0.5 * (
+        np.dot(y.T, np.linalg.solve(K, y)) + np.linalg.slogdet(K)[1] + len(y) * np.log(2 * np.pi)
+    )
     regularization = reg_param * (np.sum(beta**2))
     return -log_likelihood + regularization
 
@@ -86,7 +88,7 @@ def total_negative_log_likelihood(beta, y_all, X, kernel, reg_param=1e-6):
         Total negative log marginal likelihood.
     """
     total_log_likelihood = 0
-    for y in y_all.T:  #transposed to have shape (n_voxels, n_samples) and iterate over voxels
+    for y in y_all.T:  # transposed to have shape (n_voxels, n_samples) and iterate over voxels
         total_log_likelihood += negative_log_likelihood(beta, y, X, kernel, reg_param)
     return total_log_likelihood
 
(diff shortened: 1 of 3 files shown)


esavary:
The import of scipy in optimizers.py causes an error during the documentation build process. The issue seems to be related to the handling of dependencies or environment configuration in the Sphinx build. Has anyone encountered a similar problem, or does anyone have suggestions on how to resolve this? (Maybe @effigies @jhlegarreta )

effigies:
You can probably fix this by adding scipy to the mocked imports:

https://github.com/nipreps/eddymotion/blob/2b07b0e0e0e34b2938b9a1d34b795d3037072002/docs/conf.py#L45-L65

esavary:
> You can probably fix this by adding scipy to the mocked imports:
> 
> https://github.com/nipreps/eddymotion/blob/2b07b0e0e0e34b2938b9a1d34b795d3037072002/docs/conf.py#L45-L65

Thanks a lot, it worked!

src/eddymotion/model/optimizers.py
```diff
@@ -0,0 +1,201 @@
...
+    Log marginal likelihood function with regularization for any kernel.
+
+    Parameters
+    ----------
+    beta : array-like of shape (n_params,)
+        Log-transformed hyperparameters.
+    y : array-like of shape (n_samples,)
+        Observed data.
+    X : array-like of shape (n_samples, n_features)
+        Input data for kernel function.
```
jhlegarreta, line 38: Not sure about the convention here concerning using `array-like`  vs something more precise like
```
:obj:`list` or :obj:`~numpy.ndarray`
```

But I am fine leaving `array-like`. I usually do not tend to use the above either; I stick to only a single type, the one I provide as test data (either a list or a numpy array), even if the code may work without changes for both.

src/eddymotion/model/optimizers.py
```diff
@@ -0,0 +1,201 @@
...
+    kernel : callable
```
jhlegarreta, line 39: Rather than a callable, is this rather a `:obj:~sklearn.gaussian_process.kernels` instance?

src/eddymotion/model/optimizers.py
```diff
@@ -0,0 +1,201 @@
...
+        Kernel function which takes hyperparameters and returns a kernel instance.
+    reg_param : float, default=1e-6
```
jhlegarreta, line 41: :obj:`float`

And I would remove the default value. It is shown in the method signature, and by adding it here we run the risk of forgetting and getting it out of sync if the value ever changes in the method signature.

src/eddymotion/model/optimizers.py
```diff
@@ -0,0 +1,201 @@
...
+        Regularization parameter.
+
+    Returns
+    -------
+    float
```
jhlegarreta, line 46: Same comment as above :obj:`float`

jhlegarreta: Applies also to all methods below.

src/eddymotion/model/optimizers.py
jhlegarreta: Applies also to all methods below.

src/eddymotion/model/optimizers.py
jhlegarreta: Applies also to all methods below.

src/eddymotion/model/optimizers.py
jhlegarreta: Applies also to all methods below.

jhlegarreta (commented):
Thanks for doing this.

Made some comments about docstrings. Have not checked the details, or have not thought how this can be tested more thoroughly. @oesteban any thought on this?

If we think this is enough, it can be merged as is.

esavary:
> Thanks for doing this.
> 
> Made some comments about docstrings. Have not checked the details, or have not thought how this can be tested more thoroughly. @oesteban any thought on this?
> 
> If we think this is enough, it can be merged as is.

Thanks for your comments @jhlegarreta, I have addressed all of them in commit https://github.com/nipreps/eddymotion/pull/204/commits/34670bbacb34de1e20789a7b20de3f460a77a062

oesteban:
Superseded by #227 and other efforts:

* Log marginal likelihood: we should reuse Scikit-learn's implementation in GaussianProcessRegressor :heavy_check_mark: 
* Cross-validation: we should reuse Scikit-learn's implementation, perhaps with smac :heavy_check_mark: 
* ``stochastic_optimization_with_early_stopping`` I haven't checked whether there is a Scikit-learn implementation. If not, good to have here as a reference.

At the moment we have barely moved away from simulating one voxel and hyperparameter optimization essentially requires many voxels for it to make sense, so this PR was really ahead of its time. Very useful for reference in the future. Thanks.

PR #207, opened 2024-06-10
ENH: Outsource data filtering from estimator into its own module

oesteban:
This is the first part of  #116 sequels.


effigies (approved):
Looks good. I don't understand the sphinx warning:

The sphinx warning seems to be https://github.com/sphinx-doc/sphinx/issues/8088, which was fixed by https://github.com/sphinx-doc/sphinx/pull/8095.

Didn't dig much, I could be way off on that.

src/eddymotion/data/filtering.py
```diff
@@ -0,0 +1,94 @@
...
+    in a n-dimensional array using percentiles. It optionally enforces non-negative
+    values and scales the data to fit within a specified data type (e.g., uint8
+    for image registration). To remove outliers more robustly, the function
+    first applies a median filter to the data before calculating clipping thresholds.
+
+    Parameters
+    ----------
+    data : :obj:`~numpy.ndarray`
+        The input n-dimensional data array.
+    p_min : :obj:`float`, optional (default=35)
```
jhlegarreta, line 56: I would not put the defaults in the docstrings: if for whatever reason they are changed in the method signature, there is a risk of not updating the docstrings. I've seen this happen in other projects, frequently.

jhlegarreta (commented):
Sorry, I know I am late.

I left an inline comment. I can push a patch set if we agree that the comment is relevant.

src/eddymotion/data/filtering.py
oesteban: Yes, sounds reasonable. Please patch away :)

PR #211, opened 2024-06-13
ENH: Remove redundant `is_fitted` overrides in model classes

jhlegarreta:
Remove redundant `is_fitted` overrides in model classes.

### DIFF
--- a/src/eddymotion/model/base.py
+++ b/src/eddymotion/model/base.py
@@ -121,7 +121,6 @@ class BaseModel:
         # Keep model state
         self._model = None  # "Main" model
         self._models = None  # For parallel (chunked) execution
-        self._is_fitted = False
 
         # Setup brain mask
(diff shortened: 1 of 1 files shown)


src/eddymotion/model/base.py
```diff
@@ -121,7 +121,6 @@ def __init__(self, mask=None, **kwargs):
         # Keep model state
         self._model = None  # "Main" model
         self._models = None  # For parallel (chunked) execution
-        self._is_fitted = False
```
oesteban, line 124: Can we remove this one?

jhlegarreta: Yes: https://github.com/nipreps/eddymotion/pull/211/files#diff-a875f501910044a7d95658fb83740e2c5c6c1693e7e6808703d282441db82be8L130

PR #212, opened 2024-06-18
ENH: Make grad vec angle computation method accept two input parameters

jhlegarreta:
Make the gradient vector computation method accept two input parameters: instead of compute the angles across vectors in a single array, accept two input parameters and compute the angles across the vectors in them.

This change is introduced following the discussions on the kernel used for the Gaussian process requiring two inputs.

Take advantage of the commit to parameterize the testing and add more test cases.

Reword the comments in the testing for the sake of clearliness.

### DIFF
--- a/src/eddymotion/model/gradient_utils.py
+++ b/src/eddymotion/model/gradient_utils.py
@@ -23,7 +23,7 @@
 import numpy as np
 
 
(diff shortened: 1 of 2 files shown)


esavary (approved):
I tested it, it looks ready to merge to me.

oesteban (approved):

PR #213, opened 2024-06-22
ENH: Add single-shell covariance models

jhlegarreta:
Add the exponential and spherical covariance models for single-shell dMRI data following Andersson 2015.


oesteban:
Does this deprecate #201?

jhlegarreta:
> Does this deprecate https://github.com/nipreps/eddymotion/pull/201?

No, PR #201 will cover the multi-shell case, and this functions will be re-used as necessary there. This PR is useful as is for the single-shell case (see https://github.com/nipreps/eddymotion/pull/203/files#r1649732288), and hopefully allows to tell more easily whether up to this functions things provide the expected results.

This is ready to be merged.

PR #214, opened 2024-06-22
ENH: Add dMRI covariance model notebook

jhlegarreta:
Add dMRI covariance model notebook.

PR #215, opened 2024-07-04
ENH: Add observed covariance from HCP data to cov notebook

jhlegarreta:
Add observed covariance from HCP data to the notebook demonstrating the covariance functions.


jhlegarreta:
Following https://github.com/nipreps/eddymotion/pull/214#issuecomment-2209132845.

oesteban:
I'm getting `/data/hcp/bvals` does not exist - can we maybe set up an env variable for the path where data is?

jhlegarreta:
> I'm getting /data/hcp/bvals does not exist

Sorry, I should have mentioned in that we should try with some local HCP data that we may have. I did not use the H5 file used in other examples because an HCP participant is used in the figure.

> can we maybe set up an env variable for the path where data is?

We could. If you know some HCP participant that is up there somewhere in OpenNeuro we could use that for the notebook.

oesteban:
HCP should not be out there... I'll send you an email with solutions.

arokem:
If you have the S3 credentials from HCP set up you can use DIPY's 'fetch_hcp':

https://github.com/dipy/dipy/blob/master/dipy/data/fetcher.py#L2599

oesteban:
@jhlegarreta can you send me the dwi, bvecs, bvals and mask you are using before tomorrow's meeting?

oesteban:
@jhlegarreta thanks for the data - closing this in favor of #218 (set you as a co-author of the commit to credit your original work here, hope that's fine!)

PR #216, opened 2024-07-05
ENH: Base implementation for our GP kernels

oesteban:
Follows scikit-learn's ``PairwiseKernel`` implementation to integrate Andersson's kernels in the original eddy papers.


oesteban:
Only missing the derivative of the kernel :P -- will look at that next week.

Fixed the test with the DotProduct + WhiteKernel

Seems to be pointing at the right direction :)

src/eddymotion/model/dipy.py
```diff
@@ -330,6 +332,55 @@ def compute_spherical_covariance(
...
+
+    n_partials = len(params)
+
+    a = params.pop("a")
+    lambda_s = params.pop("lambda_s")
+
+    min_angles = theta > a
+
+    if weighting == "spherical":
+        deriv_a = (3 * theta[min_angles] / a**2) - (1.5 * (theta[min_angles] / a) ** 2) / a
```
jhlegarreta, line 370: Missing here a 2 in the denominator of the first term (2 * a**2) ?

src/eddymotion/model/dipy.py
```diff
@@ -330,6 +332,55 @@ def compute_spherical_covariance(
...
+    elif weighting == "exponential":
+        deriv_a = theta[min_angles] * a * np.exp(-theta[min_angles] / a)
```
jhlegarreta, line 372: Missing a negative sign here and a should not be a multiplying factor ?

src/eddymotion/model/dipy.py
```diff
@@ -406,14 +436,173 @@ def compute_pairwise_angles(
...
+    def __init__(
+        self,
+        weighting: str = "exponential",
+        lambda_s: float = 2.0,
+        a: float = 0.1,
+        sigma_sq: float = 1.0,
+        lambda_s_bounds: tuple[float, float] = (1e-5, 1e4),
+        a_bounds: tuple[float, float] = (1e-5, np.pi),
+        sigma_sq_bounds: tuple[float, float] = (1e-5, 1e4),
+    ):
```
jhlegarreta, line 466: Document these?

src/eddymotion/model/dipy.py
```diff
@@ -406,14 +436,173 @@ def compute_pairwise_angles(
...
+        """
+        Return the kernel K(gtab_X, gtab_Y) and optionally its gradient.
+
+        Parameters
+        ----------
+        gtab_X: :obj:`~dipy.core.gradients.GradientTable`
+            Gradient table (X)
+        gtab_Y: :obj:`~dipy.core.gradients.GradientTable`
+            Gradient table (Y, optional)
+        eval_gradient : :obj:`bool`
```
jhlegarreta, line 497: eval_gradient : :obj:`bool`, optional

src/eddymotion/model/dipy.py
```diff
@@ -406,14 +436,173 @@ def compute_pairwise_angles(
...
+    def __call__(self, gtab_X, gtab_Y=None, eval_gradient=False):
+        """
+        Return the kernel K(gtab_X, gtab_Y) and optionally its gradient.
+
+        Parameters
+        ----------
+        gtab_X: :obj:`~dipy.core.gradients.GradientTable`
+            Gradient table (X)
+        gtab_Y: :obj:`~dipy.core.gradients.GradientTable`
+            Gradient table (Y, optional)
```
jhlegarreta, line 496: gtab_Y: :obj:\`~dipy.core.gradients.GradientTable\`, optional

Wondering if you left (X) and (Y, optional) so as to identify them better with scikit-learn.

src/eddymotion/model/dipy.py
```diff
@@ -351,9 +332,58 @@ def compute_spherical_covariance(
...
+    a = params.pop("a")
+    lambda_s = params.pop("lambda_s")
+
+    min_angles = theta > a
+
+    if weighting == "spherical":
+        deriv_a = (3 * theta[min_angles] / a**2) - (1.5 * (theta[min_angles] / a) ** 2) / a
+    elif weighting == "exponential":
+        deriv_a = theta[min_angles] * a * np.exp(-theta[min_angles] / a)
+    else:
```
jhlegarreta, line 373: Also, and although this is not important now, IMO it would make sense to have these computed in separate functions, much like their integral counterparts are available in separate functions.

jhlegarreta (commented):
Some inline comment.

src/eddymotion/model/dipy.py
oesteban: Actually:

$$
\frac{\partial}{\partial x} e^{f(x)} = \frac{\partial f(x)}{\partial x} \cdot e^{f(x)}
$$

Hence:

$$
\frac{\partial}{\partial a} e^{- \frac{\theta}{a}} = \left( - \theta \frac{\partial a^{-1}}{\partial a} \right) \cdot e^{- \frac{\theta}{a}} = \frac{\theta}{a^2} \cdot e^{- \frac{\theta}{a}}
$$

```suggestion
        deriv_a = np.exp(-theta[min_angles] / a) * theta[min_angles]  / (a ** 2)
```

Correct?

src/eddymotion/model/dipy.py
oesteban: > Wondering if you left (X) and (Y, optional) so as to identify them better with scikit-learn.

Yes

src/eddymotion/model/dipy.py
oesteban: Yeah, I'm struggling with this. On the one hand, we want consistency (i.e., implement the kernel and the gradient either in separate or single functions, but both the same). On the other hand, I can't foresee more kernels being added, so it looks an overshoot to have functions for each kernel.

src/eddymotion/model/dipy.py
```diff
@@ -351,9 +332,58 @@ def compute_spherical_covariance(
     return np.where(theta <= a, 1 - 3 * (theta / a) ** 2 + 2 * (theta / a) ** 3, 0)
```
oesteban, line 332: This should be:

$$
        C(\theta) = \begin{cases}
            1 - \frac{3 \theta}{2 a} + \frac{\theta^3}{2 a^3} & \textnormal{if} \; \theta \leq a \\
            0 & \textnormal{if} \; \theta > a
        \end{cases}
$$

```suggestion
    return np.where(theta <= a, 1 - 3 * theta / (2 * a) + 2 * (theta ** 3) / (2 * (a ** 3)), 0)
```

src/eddymotion/model/dipy.py
oesteban: Thanks for forcing me to look into this :D - I believe:

$$
\frac{\partial}{\partial a}\left( 1 - \frac{3 \theta}{2 a} + \frac{\theta^3}{2 a^3} \right) = - 1.5 \cdot \theta \frac{\partial a^{-1}}{\partial a} + \frac{\theta^3}{2} \frac{\partial a^{-3}}{\partial a} = \frac{3}{2}\left( \frac{\theta}{a^2} - \frac{\theta^3}{a^4} \right)
$$

I've revised the kernel function, which I believe was wrong (sorry for the oversight when I reviewed your PR) and now this one:

```suggestion
        deriv_a = 1.5 * (theta[min_angles] / a ** 2 - theta[min_angles] ** 3 / a ** 4)
```

src/eddymotion/model/dipy.py
jhlegarreta: OK, had mistakenly took $\theta$ as the variable here. Thanks.

src/eddymotion/model/dipy.py
jhlegarreta: The plot in the notebook should give us an idea of whether the function is correct:
https://github.com/nipreps/eddymotion/blob/main/docs/notebooks/dmri_covariance.ipynb

Even if I do not see it with the above expression either :upside_down_face:.

src/eddymotion/model/dipy.py
```diff
@@ -348,12 +329,61 @@ def compute_spherical_covariance(
 
     _ensure_positive_scale(a)
 
-    return np.where(theta <= a, 1 - 3 * (theta / a) ** 2 + 2 * (theta / a) ** 3, 0)
+    return np.where(theta <= a, 1 - 3 * theta / (2 * a) + 2 * theta ** 3 / (2 * a ** 3), 0)
```
oesteban, line 332:
```suggestion
    return np.where(theta <= a, 1 - 3 * theta / (2 * a) + theta ** 3 / (2 * a ** 3), 0)
```

src/eddymotion/model/dipy.py
oesteban:
```suggestion
    ):
    """
    Initialize a kernel with pairwise angles.
    
    Parameters
    ----------
    weighting : :obj:`str`
        The type of kernel to build (either "exponential", "sphere", or "test").
    lambda_s : :obj:`float`
        The :math:`\lambda_s` hyperparameter.
    a : :obj:`float`
        Minimum angle in rads.
    sigma_sq : :obj:`float`
        Error allowed in collinear orientations.
    lambda_s_bounds : :obj:`tuple`
        Bounds for the :math:`\lambda_s` hyperparameter.
    a_bounds : :obj:`tuple`
        Bounds for the a parameter.
    sigma_sq_bounds : :obj:`tuple`
        Bounds for the error parameter.

    """
```

src/eddymotion/model/dipy.py
oesteban:
```suggestion
        eval_gradient : :obj:`bool`, optional
```

src/eddymotion/model/dipy.py
```diff
@@ -406,14 +436,173 @@ def compute_pairwise_angles(
...
+
+    def __call__(self, gtab_X, gtab_Y=None, eval_gradient=False):
+        """
+        Return the kernel K(gtab_X, gtab_Y) and optionally its gradient.
+
+        Parameters
+        ----------
+        gtab_X: :obj:`~dipy.core.gradients.GradientTable`
+            Gradient table (X)
+        gtab_Y: :obj:`~dipy.core.gradients.GradientTable`
```
oesteban, line 495:
```suggestion
        gtab_Y: :obj:`~dipy.core.gradients.GradientTable`, optional
```

src/eddymotion/model/dipy.py
```diff
@@ -348,12 +392,61 @@ def compute_spherical_covariance(
...
+
+    n_partials = len(params)
+
+    a = params.pop("a")
+    lambda_s = params.pop("lambda_s")
+
+    min_angles = theta > a
+
+    if weighting == "spherical":
+        deriv_a = 1.5 * (theta[min_angles] / a**2 - theta[min_angles] ** 3 / a**4)
```
jhlegarreta, line 433: Should be $$\frac{3\theta}{2a^2} - \frac{3\theta^3}{2a^4}$$ So missing a 1.5 multiplier to the last term.


```suggestion
        deriv_a = 1.5 * (theta[min_angles] / a**2 - 1.5 * theta[min_angles] ** 3 / a**4)
```

jhlegarreta (commented):
The covariance computation not being correct is my fault, but given that we are fixing them in this PR:
https://github.com/nipreps/eddymotion/pull/216/files#diff-fba3c6e90b496595a3cd3cdcbd5d562082cde0b4f21629d667d1a2260a91bd37R395

We should check the plots (i.e. the curve for the spherical case): https://github.com/nipreps/eddymotion/pull/216#discussion_r1670543035

Let me know if you have an OK result, because I do not get it.

src/eddymotion/model/dipy.py
oesteban: > Should be $$\frac{3\theta}{2a^2} - \frac{3\theta^3}{2a^4}$$ So missing a 1.5 multiplier to the last term.

The current implementation is $$\frac{3}{2}\left( \frac{\theta}{a^2} - \frac{\theta^3}{a^4} \right)$$, please note the parentheses around the fractional sum in the code.

src/eddymotion/model/dipy.py
oesteban: I'm getting this:

covariance-plot

We seem to have a factor of 2 somewhere, but I presume that comes from the b-value equalization.

oesteban: (notebook updated on https://github.com/nipreps/eddymotion/pull/216/commits/c05dfc05f152d6254bc4a68a7c361f0870c33414)

oesteban:
Okay, moving on - the plot looks very reasonable and I think this is in good shape. Bear with me @jhlegarreta :)

PR #217, opened 2024-07-06
ENH: Use the attenuation distance to plot the DWI signal

jhlegarreta:
Use the attenuation distance to plot the DWI signal.


jhlegarreta:
Re https://github.com/nipreps/eddymotion/pull/202#issuecomment-2211834348.

Still not right.

output_signal_distance

oesteban:
@jhlegarreta what's the status of this - should we get it over the finish line?

PR #225, opened 2024-08-30
WIP: Fast head-motion estimation for DWI

oesteban:


src/eddymotion/conftest.py
```diff
@@ -0,0 +1,54 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2021 The NiPreps Developers <nipreps@gmail.com>
```
jhlegarreta, line 4: In order to avoid C&P'ing headers that (seemingly) have the wrong year, how about using `2021-` e.g. if the copyright dates back to 2021 (or starting with the year NiPreps started)? I do not know how the copyright things (years, the copyright itself, etc.)

I know this is WIP and not meant to be reviewed yet, so I apologize.

oesteban: I think we should discuss this in one TechMon meeting because it's a recurring issue emerging across NiPreps. In fMRIPrep, I think we have dropped the year in the shebang and we only have it once in the package metadata.

That said, the year should be the date of creation of the copyrighted item, so technically, this should be the date of creation of the object - i.e., the year when the file was first created. This is hard to track, since often we upstream code from one place to another.

Let's chat about it and see what we all can agree :)

oesteban:
NOTE: Requires #230 to be merged first.

scripts/optimize_registration.py
```diff
@@ -0,0 +1,222 @@
...
+import logging
+from os import getenv
+from pathlib import Path
+from shutil import rmtree
+
+import nibabel as nb
+import nitransforms as nt
+import numpy as np
+from ConfigSpace import Configuration, ConfigurationSpace
+from smac import HyperparameterOptimizationFacade, Scenario
```
jhlegarreta, line 35: Should `ConfigSpace` and `smac` be added to the dependencies in `pyproject.toml`?

src/eddymotion/registration/ants.py
```diff
@@ -86,6 +110,217 @@ def _prepare_registration_data(dwframe, predicted, affine, vol_idx, dirname, reg
...
+    fixed_path,
+    moving_path,
+    fixedmask_path=None,
+    movingmask_path=None,
+    init_affine=None,
+    default="b0-to-b0_level0",
+    **kwargs,
+):
+    """
+    Generate an ANTs' command line.
```
jhlegarreta, line 161: Not a show stopper at this stage, but we may be willing to add parameter docstrings for this method at some point.

src/eddymotion/registration/utils.py
```diff
@@ -0,0 +1,59 @@
...
+import nibabel as nb
+import nitransforms as nt
+import numpy as np
+
+
+def displacements_within_mask(
+    maskimg: nb.spatialimages.SpatialImage,
+    test_xfm: nt.base.BaseTransform,
+    reference_xfm: nt.base.BaseTransform | None = None,
+) -> np.ndarray:
```
jhlegarreta, line 38: Document method?

src/eddymotion/registration/utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+    diffs = targets - xyz if reference_xfm is None else targets - reference_xfm.map(xyz)
+    return np.linalg.norm(diffs, axis=-1)
+
+
+def displacement_framewise(
+    img: nb.spatialimages.SpatialImage,
+    test_xfm: nt.base.BaseTransform,
+    radius: float = 50.0,
+):
```
jhlegarreta, line 54: Same as above.

env.yml
```diff
@@ -0,0 +1,28 @@
...
+    - dipy>=1.3.0
+    - h5py
+    - joblib
+    - tox
+    - tox-gh-actions
+    - coverage
+    - pytest
+    - pytest-cov
+    - pytest-env
+    - pytest-xdist
```
jhlegarreta, line 26: Unrelated to this PR, but wondering whether all dependencies in `pyproject.toml` should be listed here.

jhlegarreta (commented):
Thanks for doing this.

Except for the comment on the `ConfigSpace` and `smac` dependencies requiring to be listed in `pyproject.toml`, the rest of the comments (documentation) may be by-passed if you need this to be merged early.

jhlegarreta:
CI build is showing a warning:
```
  =============================== warnings summary ===============================
  src/eddymotion/registration/ants.py::eddymotion.registration.ants._massage_mask_path
    /home/runner/work/eddymotion/eddymotion/src/eddymotion/registration/ants.py:146: UserWarning: More mask paths than levels
      warn("More mask paths than levels", stacklevel=1)
```

https://github.com/nipreps/eddymotion/actions/runs/11114334199/job/30880581414#step:8:1012

If this test is intended, maybe the corresponding test should use a context manager to capture it, e.g.
```
 with warnings.catch_warnings():
        warnings.filterwarnings(
            "ignore",
            message="More mask paths than levels",
        )
  (...)
```

Either way, it would be good to make warnings be considered as errors (flag `--warnings-as-errors` in pytest).

jhlegarreta (dismissed):
Thanks for doing all this.

oesteban:
Waiting on #232.

src/eddymotion/registration/utils.py
```diff
@@ -0,0 +1,114 @@
...
+
+from itertools import product
+
+import nibabel as nb
+import nitransforms as nt
+import numpy as np
+
+
+def displacements_within_mask(
+    maskimg: nb.spatialimages.SpatialImage,
```
jhlegarreta, line 42: Probably easier to tell if named `mask_img`, much like `bmask_img`.

jhlegarreta, line 42: You have used ":class:`~nibabel.spatialimages.SpatialImage`" in
https://github.com/nipreps/eddymotion/pull/225/files#diff-097d3fe0cd1687fc7df869ef3ace9c3dd40ae4e9d292a97ee02d45686bbe6ce9R428

Also, if we prefer to adopt `nb.spatialimages.SpatialImage` rather than `nibabel.nifti1.Nifti1Image`, we may need to revisit other docstrings.

jhlegarreta (dismissed):
Some documentation/style comments that can be addressed in a separate PR.

src/eddymotion/registration/utils.py
jhlegarreta: Ah, yes, my bad, not awake yet. But in the docstrings sometimes you have used ":class:`~`" vs ":obj:`~`".

oesteban: :obj: is for when you don't want to break your brain and think the exact type of reference. The ~ symbol just tells sphinx to hide the full module tree and show only the name.

PR #227, opened 2024-09-25
ENH: Add GP error analysis experiment script

jhlegarreta:
Add GP error analysis experiment script.


jhlegarreta:
A few notes:
- Pardon my not very elaborated commit message.
- Started placing the methods in new modules, but I do not think that is useful: even if e.g. a crossing fiber simulation script would benefit from some methods, these methods are not required for the framework to work/not necessary for the package, they would require adding tests, etc. So unnecessary work. So I would sacrifice that for having duplicated code across script in this case. At least for now.
- The script can be launched as e.g.
```
pyhon /path/to/eddymotion/scripts/dwi_estimation_error_analysis.py \ 
  60 1000 100 \
  --evals1 0.0015 0.0003 0.0003 \
  --snr 20 \
  --repeats 3 \
  --kfold 1 3 5 15 30
```

scripts/dwi_estimation_error_analysis.py
```diff
@@ -0,0 +1,302 @@
...
+        gp_model = GaussianProcessModel(
+            kernel_model=kernel_model, lambda_s=lambda_s, a=a, sigma_sq=sigma_sq
+        )
+
+        # Create the training mask leaving out the requested number of samples
+        train_mask = create_random_train_mask(gtab, n)
+
+        # Simulate the fitting a number of times: every time the signal created will be a little
+        # different
+        for _ in range(repeats):
```
jhlegarreta, line 167: One thing about these repeats is that the GP itself is able to give us a std dev,
https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html#sklearn.gaussian_process.GaussianProcessRegressor.predict

but we are not using it,
https://github.com/nipreps/eddymotion/blob/fcd9e94e02f706ed0e2918a5d7e6896825fec563/src/eddymotion/model/_dipy.py#L141

so we may want to avoid an additional parameter and loop and use the former.

jhlegarreta:
Using the std deviation value of scikit-learn's GP regressor class (committed in 
ea93c4a for the records, but most likely will drop when we get this sorted out) we get the below results, which is again not what we would expect

jhlegarreta:
@oesteban Adopted `scikit learn`'s `KFold` to create folds in acf683b. One would call the script as e.g.
```
pyhon /path/to/eddymotion/scripts/dwi_estimation_error_analysis.py \
  60 1000 100 --evals1 0.0015 0.0003 0.0003 --repeats 3 --snr 20 --kfold 3 5 7 10 20 120
```

- Kept the `repeats` parameter but not using them now.
- The output is not what I would expect: error and std dev are constant + the RMSE with SNR 20 is smaller than with SNR 10

oesteban:
@jhlegarreta I've moved things to use scikit-learn's cross-validation to avoid errors on our side implementing the CV framework:

1. Updated the model to take arrays instead of gradient tables (https://github.com/nipreps/eddymotion/pull/227/commits/ff08c26660ed6c3cb5f29a0c6cfd37e5677ceaf1). We probably will want to add a checkpoint in fit and check if `y` is a gradient table (and extract bvecs accordingly). But ideally, we want the model to self adapt to incoming ndarrays or DIPY's gradient tables.
2. Then, I use the default k-fold CV (https://github.com/nipreps/eddymotion/pull/227/commits/b963b968c8915f2107963fb7b06c54d7c0cfc855).

Running this:

```
python scripts/dwi_estimation_error_analysis.py 60 1000 100 --evals1 0.0015 0.0003 0.0003 --repeats 3 --snr 20 --kfold 5 10 60 --repeats 200
```
I get
```
{5: (-6.5451513798923475, 0.7829974075194627), 10: (-6.704245824324263, 1.451348158720791), 60: (-6.339094677419443, 2.883843582961101)}
```

which I believe is consistent with the expectations:
- RMSE is negative because scikit-learn negates it.
- Error is constant around 6. something, however the spread grows with larger folds. This makes sense to me because the more folds the more fits and the bigger the probability of getting RMSEs further from the average. Also the average is with only two values, so it's also easier to get larger errors. With repeats=200 then k=5 means 1000 fits while k=60 means 12000 fits. I'll try to compensate for this later.

Regarding the number of directions in training, the interpretation of the k-fold is also different from what I think we were discussing:

k = 5 -> 5 folds, 120/5=24 bvecs per test fold -> 96 bvecs for training.
k = 60 -> 60 folds, 120/60 = 2 bvecs per test fold -> 118 bvecs for training.

I keep receiving the following warnings though:

```
/data/home/oesteban/workspace/scikit-learn/sklearn/gaussian_process/kernels.py:442: ConvergenceWarning: The optimal value found for dimension 0 of parameter a is close to the specified lower bound 1e-05. Decreasing the bound and calling fit again may find a better value.
/data/home/oesteban/workspace/scikit-learn/sklearn/gaussian_process/kernels.py:452: ConvergenceWarning: The optimal value found for dimension 0 of parameter lambda_s is close to the specified upper bound 10000.0. Increasing the bound and calling fit again may find a better value.
/data/home/oesteban/workspace/scikit-learn/sklearn/gaussian_process/kernels.py:442: ConvergenceWarning: The optimal value found for dimension 0 of parameter sigma_sq is close to the specified lower bound 1e-05. Decreasing the bound and calling fit again may find a better value.
```
and:
```
/data/home/oesteban/workspace/scikit-learn/sklearn/gaussian_process/_gpr.py:659: ConvergenceWarning: lbfgs failed to converge (status=2):
ABNORMAL_TERMINATION_IN_LNSRCH.

Increase the number of iterations (max_iter) or scale the data as shown in:
    https://scikit-learn.org/stable/modules/preprocessing.html
  _check_optimize_result("lbfgs", opt_res)
```

WDYT?

oesteban:
Decreasing the SNR did not help with the warnings. My hypothesis being that noisier data would regularize the problem.

oesteban:
Okay, decreasing the SNR had the expected effect on the cross-validation:

```
python scripts/dwi_estimation_error_analysis.py 60 1000 100 --evals1 0.0015 0.0003 0.0003 --repeats 3 --snr 10 --kfold 5 10 60 --repeats 200
```

Results in

```
{5: (-13.634124763730938, 1.0844600539135052), 10: (-13.739925211912865, 1.6677184190441654), 60: (-13.892240109267137, 4.902059389908273)}
```

jhlegarreta:
31e5972 adds the capability to generate a carpet plot with the ground truth and the estimated signal (see an example in https://github.com/nipreps/nireports/issues/137) and a correlation plot. The calls are commented since after pulling the most recent commits I've seen that we no longer have the data available; had done this prior to your latest commits.

oesteban:
@jhlegarreta I've found several issues - see https://github.com/nipreps/eddymotion/pull/227/commits/10a576d7459c8e52e2588b7666d109d46dba8b54

I'll keep digging the kernel.theta setting. Keep you posted.

oesteban:
@jhlegarreta Progress https://github.com/scikit-learn/scikit-learn/pull/30098

jhlegarreta:
Thanks for all this investigation.

Let me know if you are able to push the latest changes so that I can continue investigating on my end.

---
Unrelated note: eventually, after we manage to fix the current issues and if it makes sense to bring the implementation closer to DIPY's reconstruction models, we may want to use DIPY's cross validation:
https://github.com/dipy/dipy/blob/1961a91e2130aa9e2891a5610a79e2c9aca24d88/dipy/reconst/cross_validation.py#L58
https://github.com/dipy/dipy/blob/1961a91e2130aa9e2891a5610a79e2c9aca24d88/doc/examples/kfold_xval.py

scripts/dwi_estimation_error_analysis.py
```diff
@@ -0,0 +1,642 @@
...
+    for n in kfold:
+        # Assumptions:
+        #  - Consecutive indices in the folds
+        #  - A single b0
+        kf = KFold(n_splits=n, shuffle=False)
+
+        # Define the Gaussian process model instance
+        gp_model = EddyMotionGPR(
+            kernel=SphericalKriging(a=a, lambda_s=lambda_s),
+            alpha=sigma_sq,
```
oesteban, line 331: Our $\sigma^2$ corresponds to $\alpha$ in Scikit-learn's implementation by comparison of Andersson 2015 and Rasmussen 2006 (https://gaussianprocess.org/gpml/chapters/RW.pdf).

scripts/dwi_estimation_error_analysis.py
```diff
@@ -0,0 +1,642 @@
...
+    gpm = EddyMotionGPR(
+        kernel=SphericalKriging(a=2.15, lambda_s=120),
+        alpha=50,
+        optimizer=None,
+    )
+
+    X = gtab[~gtab.b0s_mask].bvecs
+    y = signal[~gtab.b0s_mask]
+
+    rkf = RepeatedKFold(n_splits=cv, n_repeats=120 // cv)
```
oesteban, line 398: Use a repeated k-fold so that 5-fold and 60-fold are somewhat comparable.

scripts/dwi_estimation_error_analysis.py
```diff
@@ -0,0 +1,642 @@
...
+        for n in args.kfold
+        for _ in range(args.repeats)
+    }
+
+    print({n: (np.mean(scores[n]), np.std(scores[n])) for n in args.kfold})
+
+    # Compute the error
+    # rmse, std_dev = compute_error(data, args.repeats, args.kfold)
+
+    # Plot
```
oesteban, line 609: We'll want to reenable plotting before merging.

src/eddymotion/model/_dipy.py
```diff
@@ -176,32 +114,30 @@ def __init__(
...
+        self.kernel = KernelType(
+            a=a,
+            lambda_s=lambda_s,
         )
 
     def fit(
         self,
         data: np.ndarray,
-        gtab: GradientTable | None = None,
+        gtab: GradientTable | np.ndarray,
```
oesteban, line 129: With the separation of `_dipy` and `_sklearn`, these types use GradientTables and arrays. Under `_sklearn` only arrays are allowed.

src/eddymotion/model/_sklearn.py
```diff
@@ -0,0 +1,435 @@
...
+from scipy import optimize
+from scipy.optimize._minimize import Bounds
+from sklearn.gaussian_process import GaussianProcessRegressor
+from sklearn.gaussian_process.kernels import (
+    Hyperparameter,
+    Kernel,
+)
+from sklearn.metrics.pairwise import cosine_similarity
+
+BOUNDS_A: tuple[float, float] = (1e-4, np.pi)
```
oesteban, line 102: $a$ is an angular distance threshold, so boundaries should be (0, $\pi$]

src/eddymotion/model/_sklearn.py
```diff
@@ -0,0 +1,435 @@
...
+
+    References
+    ----------
+    .. [Andersson15] J. L. R. Andersson. et al., An integrated approach to
+       correction for off-resonance effects and subject movement in diffusion MR
+       imaging, NeuroImage 125 (2016) 1063-11078
+
+    """
+
+    cosines = np.clip(cosine_similarity(X, Y, dense_output=dense_output), -1.0, 1.0)
```
oesteban, line 434: Use scikit-learn's cosine similarity, as it is very fast and code is clearer.

oesteban:
Currently, I have:

```
python scripts/dwi_estimation_error_analysis.py 60 1000 100 --evals1 0.0015 0.0003 0.0003 --snr 20 --kfold 3 5 10 15 20 25 30 40 50 60 --repeats 1000
{3: (-5.323663473768337, 0.5078300442425429), 5: (-5.592537291455177, 0.6344872918456032), 10: (-6.457823460261486, 1.1560694555678308), 15: (-6.0008898214350985, 1.2415790700142275), 20: (-4.785889809484692, 1.2121841573793941), 25: (-5.02350508651579, 1.7419826969482066), 30: (-4.902673405605454, 1.6647001726929596), 40: (-4.886979705096015, 1.9020697112634937), 50: (-5.276554893079895, 2.636081796110276), 60: (-5.393546086268717, 2.5409755115661166)}
```

scripts/dwi_estimation_error_analysis.py
```diff
@@ -0,0 +1,207 @@
...
+        print(f"Finished {n}-fold cross-validation")
+    
+    scores_df = pd.DataFrame(scores)
+    scores_df.to_csv("cv_scores.tsv", sep="\t", index=None, na_rep="n/a")
+
+    grouped = scores_df.groupby(["n_folds"])
+    print(grouped[["rmse"]].mean())
+    print(grouped[["rmse"]].std())
+
+    # Plot - import plot_* from eddymotion.viz.signals
```
oesteban, line 174: Keeping the code here for future. Happy to wait for these plots to happen, but also ready to merge and work the visualizations elsewhere.

scripts/dwi_estimation_error_analysis.py
oesteban: That's the stdev of the prediction, it's not of the error. It serves a different purpose.

oesteban:
Running a comparison right now - let's keep moving :)

PR #228, opened 2024-09-26
ENH: Add a script to plot the signal estimated by the GP

jhlegarreta:
- ENH: Add a script to plot the signal estimated by the GP
- STYLE: Rename the error analysis script to honor better its purpose
- STYLE: Rename the `EddyMotionGPR` instance to honor better its classname


scripts/dwi_estimation_plot.py
```diff
@@ -0,0 +1,331 @@
...
+    # Predict on an oversampled set of random directions over the unit sphere
+    # theta, phi = create_random_polar_coordinates(SAMPLING_DIRECTIONS, seed=seed)
+    # sph = Sphere(theta=theta, phi=phi)
+
+    # ToDo
+    # Not sure why all predictions are zero in gpfit.predict(sph.vertices)
+    # Also, when creating the convex hull, the gtab required is the one that
+    # would correspond to the new directions, so a new gtab would need to be
+    # generated
+    # return signal[~gtab.b0s_mask], gpfit.predict(sph.vertices)
```
jhlegarreta, line 177: Not sure why this does not work (predictions are all 0's).

jhlegarreta:
```
pyhon /path/to/eddymotion/scripts/dwi_estimation_plot.py \ 
  60 1000 100 \
  --evals1 0.0015 0.0003 0.0003 \
  --snr 20
```

gives the below result (with the caveat mentioned in https://github.com/nipreps/eddymotion/pull/228#discussion_r1776148824)

scripts/dwi_estimation_plot.py
jhlegarreta: Some investigation here:
- If I concatenate the desired 200 directions to the gtab directions used for the fitting and predict:
  ```
  _bvecs = np.vstack([gtab.bvecs[~gtab.b0s_mask], sph.vertices])
  gpfit.predict(_bvecs)
  ```
  it still results in all predictions corresponding to `sph.vertices` being 0.

  The issue arises from https://github.com/scikit-learn/scikit-learn/blob/f6871e80b1d11dfd18d3b85a9f7799c56371cb64/sklearn/gaussian_process/_gpr.py#L441. At that point the values in excess of the training mask are all zero. In the immediately previous line, the covariance matrix `K_trans = self.kernel_(self.X, self.X_train_)` the diagonal contains non-zero values (all valued 1e4: looks like the `sigma_sq_bounds` value).

  If I predict on `sph.vertices` (`gpfit.predict(sph.vertices)`), all entries in the covariance matrix are zero.

- If I keep 60 dirs ad the script input parameter to create the gradient table, and request less directions (`SAMPLING_DIRECTIONS = 20`, or `100`), `K_trans` is all zeros (largest value is 1e-62). 

 - Using 20, 30, 40 directions too create the gradient table, and predicting on `SAMPLING_DIRECTIONS = 100` does not result in zeros, but if I use 45 or 50 gradient table directions the prediction is all in zeros and `K_trans` is all zeros. Using 42 directions to create the gradient table, I already get values in the order of 1e-6 in the predictions.

   Same happens if I try to predict `SAMPLING_DIRECTIONS = 20` and use 42 directions for the gradient table: I get 1e-6 values in the prediction.

- Now if I set the parameter `a = 10` I get some non-zero predictions in `y_mean`. Have not checked the result in the plot.

So, this looks related to the kernel and the hyperparameter optimization ?

Additional not sure if the conditioning of the problem has anything to do with this.

jhlegarreta:
After 9358298 the script is able to simulate accept one, or two or three crossing fibers, e.g.

One fiber:
```
pyhon /path/to/eddymotion/scripts/dwi_estimation_plot.py \ 
  60 1000 100 \
  --angles 0 0 \
  --evals 0.0015 0.0003 0.0003 \
  --snr 20
```
Two fibers:
```
pyhon /path/to/eddymotion/scripts/dwi_estimation_plot.py \ 
  60 1000 100 \ 
  --angles 0 0 90 0 \
  --evals 0.0015 0.0003 0.0003 0.0015 0.0003 0.0003 \
  --snr 20
```
Three fibers:
```
pyhon /path/to/eddymotion/scripts/dwi_estimation_plot.py \ 
  60 1000 100 \
  --angles 0 0 90 0 90 90 \
  --evals 0.0015 0.0003 0.0003 0.0015 0.0003 0.0003 0.0015 0.0003 0.0003 \
  --snr 20
```

jhlegarreta:
A few notes:
- I have tested the error analysis and plot scripts and they work as intended (or at least as they were working prior to this commit). 
  - The std dev is monotonically increasing only after the 10-fold CV in my experiments (happens in the `main` branch as well: @oesteban you may want to check this; if we conclude that there is something wrong, I'd rather try to fix that in a separate PR).
- Checking the overlap between/merging `eddymotion.testing.simulations.create_random_polar_coordinates` and `eddymotion.testing.simulations.create_random_polar_angles` is left for a separate PR.

PR #230, opened 2024-09-30
MAINT: Revision of GHA migration

oesteban:
Revises the GHA migration to make sure that tests run:

- Added dataset download using datalad (as it was the case in Circle)
- Improved the tox config to ensure environment variables are considered
- Added execution of pytest on the `test/` folder.
- Does not test on a matrix of pythons because the python version should not introduce differences (to consider when we have code where it could be a problem).

IMHO, the addition of tox deviates from other projects (and as far as I understand, we were not planning on using tox) and has introduced some complexity. In particular, I had to set one test to skip because of (I suspect) some minimal environment change tox introduces.

For now, I wanted to focus on the download of data and execution of all tests. Future PRs can address the removal of tox if we finally decide against it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,7 @@ defaults:
 env:
   FORCE_COLOR: true
   TEST_DATA_HOME: /home/runner/eddymotion-tests/
+  ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS: 4
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


jhlegarreta (approved):
Thanks for doing this.

I apologize for not having inspected closely the CircleCI config file, and for inadvertently discarding the data download and testing the `test/` folder scripts. Now I realize that the CI duration should have been indicative of this.

Sorry if the adoption of tox made things more complicated; assumed there was some consensus as it had been adopted on other NiPreps repositories. Can revisit later.

Merging for not to go forward.

PR #232, opened 2024-10-01
MAINT: Revert addition of tox

oesteban:
Let's see whether the failing test comes from the tox insulation of the environment.


effigies:
The nitransforms bug was related to adding per volume transforms to a chain. Should be fixed in the latest release. Are you able to fix this issue by pinning <24?

oesteban:
> Are you able to fix this issue by pinning <24?

That was about to be my next commit :D

This is where I think we are hitting game over:

https://github.com/nipreps/eddymotion/blob/4a6cf72c5fa3dc905215c9db4e32b9298a3cebb1/test/test_integration.py#L44-L55

oesteban:
It is nitransforms.

PR #236, opened 2024-10-08
MAINT: Move test data files to a dedicated folder inside `test` folder

### LINKED ISSUE #233: Clean up `data` module
### What would you like to see added in this software?

The `data` module contains both Python code and actual data. For the sake of cleanliness, I think it would be best to separate these. If we cannot transfer the actual data to a separate repository (e.g. OSF or other), we may want to create a `data` folder one level above the current one to host actual data files, leaving the Python `data` module where it currently dwells.

### Do you have any interest in helping implement the feature?

Yes

### Additional information / screenshots

_No response_

jhlegarreta:
Move test data files to a dedicated folder inside the `test` folder, and edit accordingly the `_datadir` path used across doctests and examples.

Fixes #233.


oesteban:
These are test-data files, would it make sense to move them into `/test/` instead?

jhlegarreta:
> These are test-data files, would it make sense to move them into /test/ instead?

I don't have a strong preference, but in order to keep the code separated from the data, if we move these files to the `test` folder, I would move them to `test/data`. Let me know if you feel comfortable with this. Thanks.

oesteban:
Sorry, yes, /test/data was the intended proposal

PR #237, opened 2024-10-23
ENH: Refactor of GP regression models

oesteban:
Spun off of #227.


oesteban:
I'm positive this is necessary. Let's move forward so that #227 can be merged soon too.

PR #244, opened 2024-10-24
ENH: Improving GP models

oesteban:


oesteban:
@jhlegarreta our simulated data may not be realistic enough for our tests:

synthetic-covariance-plot

That said, convergence issues also happen when using b=3000 of HCP.

oesteban:
@jhlegarreta @yasseraleman -- the new plot accounting for the normalization of the data differently (see my commit message in 3136256):

covariance-plot

Perhaps the optimizer is moving *a* towards $\frac{\pi}{2}$ because it is indeed a better fit than *a* = 1.23.

jhlegarreta:
> @jhlegarreta @yasseraleman -- the new plot accounting for the normalization of the data differently (see my commit message in https://github.com/nipreps/eddymotion/commit/31362568767773b94c780abbe40e5a793ecef43a):

@oesteban :tada: looks reasonable.

>  It is still unclear to me how :math:`f^{*}` is interpolated from :math:`\hat{\mathbf{f}}` (grand mean?), or whether the fitted model accounts for that factor (I doubt it).

Maybe I'm being very naive, but they assume that $f^{*}$ is the target value, and simply add $\hat{f}$ once they're done with the estimation following the text? That said, I have not found where in the *eddy* folks they add back the mean:
https://git.fmrib.ox.ac.uk/fsl/eddy/-/blob/2480dda293d4cec83014454db3a193b87921f6b0/DiffusionGP.cpp#L252

Maybe since they add it back to the predictions when writing the volumes as they have kept track of the value. But have not found the place, unless it is this one:
https://git.fmrib.ox.ac.uk/fsl/eddy/-/blob/2480dda293d4cec83014454db3a193b87921f6b0/DiffusionGP.cpp#L132

oesteban:
> Maybe I'm being very naive, but they assume that $f^{*}$ is the target value, and simply add $\hat{f}$ once they're done with the estimation following the text? That said, I have not found where in the *eddy* folks they add back the mean

I don't think that's naive and I understand the same of the text. But then the question is you can obtain $\bar{\mathbf{f}}$ by just calculating the mean along the voxels axis (obtaining one mean per b-vector). However, in prediction, you don't know the mean of the orientations. So perhaps eddy is kind of *cheating* and getting the parameter from the testing data (fair enough, which would explain why you didn't find it in the prediction part of the code) or they are actually interpolating $\bar{\mathbf{f}}$ on the surface (which doesn't seem to be implemented within eddy).

oesteban:
Addendum: This would invalidate our cross-validation unless we regress data to $\bar{\mathbf{f}}$ before entering cross-validation.

src/eddymotion/model/_sklearn.py
```diff
@@ -98,98 +99,225 @@
...
+
+        y : array-like of shape (n_samples,) or (n_samples, n_targets)
+            Target values.
+
+        Returns
+        -------
+        :obj:`~sklearn.gaussian_process.GaussianProcessRegressor`
+            GaussianProcessRegressor class instance.
+        """
+        super().fit(X, y - y.mean(1)[:, None])
```
jhlegarreta, line 279: Related to the comment in https://github.com/nipreps/eddymotion/pull/244#issuecomment-2439882062, 

> (...) calculating the mean along the voxels axis obtaining one mean per b-vector) (...)

and the https://github.com/nipreps/eddymotion/commit/31362568767773b94c780abbe40e5a793ecef43a commit message

> They certainly removed the mean (I don't know if gradient-wise mean or the grand mean),

I'm more inclined to think that the mean should be computed over the b-vecs instead of the voxels, as our `x` are the b-vecs and the data lives on the sphere. But from their `mean_correct` method `m` seems to index each gradient volume, and `s` seems to index each voxel, so it looks like they are averaging over voxels. :pensive:.

oesteban:
Improved:
covariance-plot

PR #246, opened 2024-10-25
STYLE: Prefer using "Copyright" over the symbol in file headers

jhlegarreta:
Prefer using "Copyright" over the symbol in file headers.


jhlegarreta:
Not sure if this is a consequence of the IDE formatting capabilities when C&Ping. Else, if the symbol is preferred, then I'll change the rest of the files and the `NOTICE` in the repository root.

oesteban:
It's okay either way. Let's start with what is the standard now.

PR #249, opened 2024-10-29
ENH: Improve CLI of the dwi/error experiment

oesteban:
Moves to optional many arguments so that the script is easier to call (previously, you needed to check the right ordering always and positional parameters are very easy to mix-up).

It also makes optional several features: writing the generated synthetic data, writing the predictions, and the k-fold cross validation (added a single-run possibility useful for debugging).

### DIFF
--- a/scripts/dwi_gp_estimation_error_analysis.py
+++ b/scripts/dwi_gp_estimation_error_analysis.py
@@ -211,6 +211,7 @@ def main() -> None:
                 scores["rmse"] += cv_scores.tolist()
                 scores["repeat"] += [i] * len(cv_scores)
                 scores["n_folds"] += [n] * len(cv_scores)
+                scores["bval"] += [args.bval-shell] * len(cv_scores)
                 scores["snr"] += [snr_str] * len(cv_scores)
 
             print(f"Finished {n}-fold cross-validation")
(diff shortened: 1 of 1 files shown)


scripts/dwi_gp_estimation_error_analysis.py
```diff
@@ -164,25 +202,34 @@ def main() -> None:
...
-    print(grouped[["rmse"]].std())
+    if args.kfold:
+        # Use Scikit-learn cross validation
+        scores = defaultdict(list, {})
+        for n in args.kfold:
+            for i in range(args.repeats):
+                cv_scores = -1.0 * cross_validate(X, y.T, n, gpr)
+                scores["rmse"] += cv_scores.tolist()
+                scores["repeat"] += [i] * len(cv_scores)
+                scores["n_folds"] += [n] * len(cv_scores)
```
jhlegarreta, line 213:
```suggestion
                scores["n_folds"] += [n] * len(cv_scores)
                scores["bval"] += [args.bval-shell] * len(cv_scores)
```

I think serializing the b-value is relevant. Revisiting Andersson2015, I've gone further down and have seen Fig 8 (so similar to what we had discussed, now adding different b-values). So I'd recover that.

jhlegarreta (dismissed):
An in-line suggestion.

PR #255, opened 2024-11-05
ENH: Add GP estimation notebook

jhlegarreta:
Add GP estimation notebook.


jhlegarreta:
A few notes:
- In my CV I am only allowing two values; I've set it to 5 locally, but takes a while, so for now leaving 2 in the notebook.
- Andersson2015 use lambda [119, 1.15, 186] (p. 175) for [lambda, a, sigma^2]. The values I've used are those in https://github.com/nipreps/eddymotion/blob/main/docs/notebooks/dmri_covariance.ipynb. I have tried with lambda 100 without an improvement in results.

jhlegarreta:
For the records, the issues I had when computing the GP and fixed by 03b3e54, they were due to these two lines arranging the data in a different way:

```
dwi_mask = np.repeat(brain_mask[..., np.newaxis], shell_data.shape[-1], axis=-1)
y = shell_data[dwi_mask]
y_train = y.reshape((X_train.shape[0], -1))
```

wrt to:
```
y = shell_data[brain_mask]
y_train = y.T
```

(in the example above I am pretending all data is used for the GP fit to remove any effect of the DWI train_test_mask)

Thanks Óscar, for having taken another, simpler approach in 03b3e54 that allowed to get around the issue.

Ready to be merged.

oesteban (approved):

PR #257, opened 2024-11-08
ENH: Run Jupyter notebooks in the `tox` testing step

jhlegarreta:
Run Jupyter notebooks in the `tox` testing step.

Add `nbconvert` and `nbclient` to the dependencies of the `test` table in `pyproject.toml` so that the notebooks can be run.


jhlegarreta:
@effigies @oesteban sorry to tag you. I'd be interested in knowing whether you'd add a separate step to `tox.ini` for this or whether you think this is OK.

effigies:
What's the old and new runtime? If they're long running, I'd suggest a new tox environment. If it's not a significant addition, this seems fine.

jhlegarreta:
> What's the old and new runtime? If they're long running, I'd suggest a new tox environment. If it's not a significant addition, this seems fine.

I see; before it was around 6 minutes, but after reading your comment I believe that it is better to add a new environment event if the running time is not that long. I am missing something, though, as the notebook environment is not being run. In fact, none of the other environments are being run either.

PR #260, opened 2024-12-08
STYLE: Make Kriging hyperparameter property naming consistent

jhlegarreta:
Make Kriging hyperparameter property naming consistent: use `hyperparameter_l` instead of `hyperparameter_beta_l` for the `beta_l` hyperparameter, much like `hyperparameter_a` is being used for the `beta_a` hyperparameter.

### DIFF
--- a/src/eddymotion/model/gpr.py
+++ b/src/eddymotion/model/gpr.py
@@ -288,7 +288,7 @@ class ExponentialKriging(Kernel):
         return Hyperparameter("beta_a", "numeric", self.a_bounds)
 
     @property
-    def hyperparameter_beta_l(self) -> Hyperparameter:
+    def hyperparameter_l(self) -> Hyperparameter:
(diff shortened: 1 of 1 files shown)


effigies:
Does anything depend on these? This is an API change, but I don't know the scope of the impact.

jhlegarreta:
> Does anything depend on these?

I would like @oesteban to review this and check the https://github.com/nipreps/eddymotion/blob/main/docs/notebooks/dwi_gp_estimation.ipynb notebook. I cannot check it myself because the data is yet to be shared in the repository. It was found out by Oscar that `scikit-learn` follows an alphabetical order when processing the kernel hyperparameters/parameters and thus he added the `beta_` prefix.

> This is an API change, but I don't know the scope of the impact.

We had not released a version with the previous API so I believe we are allowed to break the API at this stage.

oesteban (approved):

oesteban (approved):

PR #261, opened 2024-12-08
ENH: Implement squared exponential covariance function

jhlegarreta:
Implement squared exponential covariance function.


oesteban:
I believe we want to leverage sklearn's composition of kernels on this one (and kill to birds with one stone)

PR #262, opened 2024-12-08
DOC: Include special members in documentation

jhlegarreta:
Include the `__call__` and `__len__` special members when generating the documentation.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -234,6 +234,13 @@ apidoc_excluded_paths = ["conftest.py", "*/tests/*", "tests/*", "config/*"]
 apidoc_separate_modules = True
 apidoc_extra_args = ["--module-first", "-d 1", "-T"]
 
+
+# -- Options for autodoc extension -------------------------------------------
+autodoc_default_options = {
+    "special-members": "__call__, __len__",
+}
+
+
 # -- Options for intersphinx extension ---------------------------------------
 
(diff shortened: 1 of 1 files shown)


jhlegarreta:
@effigies I have been unsuccessful at adding the `__init__` documentation to the class documentation, e.g.
https://scikit-learn.org/1.5/modules/generated/sklearn.gaussian_process.kernels.RBF.html#rbf

The `autoclass_content` `both` option:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autoclass_content

does only part of the job that I'm trying to replicate here: the `__init__` contents are added to the class documentation,

eddymotion_init_vs_class_documentation

but as shown above, the arguments are not being added to the class signature (`*args: Any, **kwargs: Any` exist in the [online documentation](https://www.nipreps.org/eddymotion/main/api/eddymotion.model.gpr.html#eddymotion.model.gpr.ExponentialKriging), and I have not found a way to remove them). Moving the parameters to the class documentation from the `__init__` method and trying all three possibilities with `autoclass_content` does not do the trick.

Any suggestion ?

jhlegarreta:
Re: https://github.com/nipreps/eddymotion/pull/262#issuecomment-2525373164 PR #263 should do the trick.

oesteban (approved):

